PR #19, opened 2019-09-20
feat: flood ping module

ankitjena:
Added support for flood ping in `ping` module.

### DIFF
--- a/src/lib/filters/scraps/scrapPing.go
+++ b/src/lib/filters/scraps/scrapPing.go
@@ -35,3 +35,16 @@ func CLIPingScrap(s *string) (a *TypePingScrap) {
 	}
 	return
 }
+
+// CLIFLoodPingScrap returns packet loss
+func CLIFLoodPingScrap(s *string) (a uint64) {
+	arr := strings.Split(*s, "\n")
+	sentence := arr[3]
+	words := strings.Split(sentence, ", ")
+	percentage := strings.Split(words[2], " ")[0]
+	a, err := strconv.ParseUint(strings.TrimRight(percentage, "%"), 10, 32)
+	if err != nil {
+		panic(err)
+	}
+	return
+}
--- a/src/lib/filters/scraps/scrapPing.go
+++ b/src/lib/filters/scraps/scrapPing.go
@@ -37,14 +37,21 @@ func CLIPingScrap(s *string) (a *TypePingScrap) {
 }
 
 // CLIFLoodPingScrap returns packet loss
-func CLIFLoodPingScrap(s *string) (a uint64) {
+func CLIFLoodPingScrap(s *string) (a float64, b *TypePingScrap) {
 	arr := strings.Split(*s, "\n")
 	sentence := arr[3]
 	words := strings.Split(sentence, ", ")
 	percentage := strings.Split(words[2], " ")[0]
-	a, err := strconv.ParseUint(strings.TrimRight(percentage, "%"), 10, 32)
-	if err != nil {
-		panic(err)
+	a = strToFloat64(strings.TrimRight(percentage, "%"))
+
+	pingValuesString := strings.Split(arr[4], " ")[3]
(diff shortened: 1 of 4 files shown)


src/lib/filters/scraps/scrapPing.go
```diff
@@ -26,12 +26,25 @@ func CLIPingScrap(s *string) (a *TypePingScrap) {
...
 			}
 		}
 	}
 	return
-}
\ No newline at end of file
+}
+
+// CLIFLoodPingScrap returns packet loss
+func CLIFLoodPingScrap(s *string) (a uint64) {
```
Harkishen-Singh: `a` should be of type `float32` since the packet loss us always of a decimal value. Please change it accordingly in all respective places

src/lib/filters/scraps/scrapPing.go
```diff
@@ -26,12 +26,25 @@ func CLIPingScrap(s *string) (a *TypePingScrap) {
...
+	arr := strings.Split(*s, "\n")
+	sentence := arr[3]
+	words := strings.Split(sentence, ", ")
+	percentage := strings.Split(words[2], " ")[0]
+	a, err := strconv.ParseUint(strings.TrimRight(percentage, "%"), 10, 32)
```
Harkishen-Singh: it would be really helpful if we could get ***min/avg/max/mdev*** values of ping along with the packet loss rather than just the packet loss. This is due to the fact that sending data in a flooded environment will have a significant impact on the ping as well, thus denoting a real-world scenario. Please change this accordingly.

src/lib/modules/ping/ping_test.go
```diff
@@ -25,4 +27,15 @@ func TestHandlerPing(t *testing.T) {
...
 		}
 	}
-}
\ No newline at end of file
+}
+
+func TestHandlerFloodPing(t *testing.T) {
+	for _, inst := range urls {
+		a := HandleFloodPing(&inst, 1000)
+		if reflect.TypeOf(a).String() != "uint64" {
```
Harkishen-Singh: please make the error condition wider. like, after implementing min/avg/max/mdev in the flood ping, it would be good to check if the returned value is abnormal or not.

ankitjena: What would be abnormal in this case?

Harkishen-Singh: might be like a ping being 0ms, ping above 1000ms, or some negative integer or a string being passed on due to wrong scrapping. This would be necessarily helpful if we have some contributor which would be involved in improving this module and has by mistake messed up with the scrapping algo. 

What do you think? @ankitjena

ankitjena: Yes, that sounds good.

ankitjena:
> please fix conflicts and also remove comments @ankitjena

Which comments?

src/lib/modules/ping/ping_test.go
```diff
@@ -1,28 +1,48 @@
...
 )
 
-func TestHandlerPing(t *testing.T) {
+// func TestHandlerPing(t *testing.T) {
+// 	for _, inst := range urls {
+// 		a := HandlePing(&inst, 20)
+// 		if *a == (scrap.TypePingScrap{}) {
+// 			t.Errorf("invalid response from HandlePing")
+// 		} else {
+// 			log.Println(*a)
```
Harkishen-Singh: these

src/lib/modules/ping/ping_test.go
```diff
@@ -1,28 +1,48 @@
...
+	scrap "github.com/zairza-cetb/bench-routes/src/lib/filters/scraps"
 )
 
 var (
 	urls = []string{
 		"google.co.in",
 		"facebook.com",
 		"yahoo.com",
-		"github.com",
+		// "github.com",
```
Harkishen-Singh: this too

PR #27, opened 2019-09-25
feat: getPositionalPointerNormalized method

ankitjena:
Signed-off-by: ankitjena <ankitjena13@gmail.com>

Add getPositionalPointerMethod and tests.


test-files/loadFromStorage_testdata/test1.json
```diff
@@ -1 +1 @@
...
...019-09-17T13:00:20.450146794+05:30"},{"datapoint":8800,"normalizedTime":1568705508,"timestamp":"2019-09-17T13:00:20.450154523+05:30"},{"datapoint":8900,"normalizedTime":1568705509,"timestamp":"2019-09-17T13:00:20.450161588+05:30"},{"datapoint":9000,"normalizedTime":1568705510,"timestamp":"2019-09-17T13:00:20.450168603+05:30"},{"datapoint":9100,"normalizedTime":1568705511,"timestamp":"2019-09-17T13:00:20.45017554+05:30"},{"datapoint":9200,"normalizedTime":1568705512,"timestamp":"2019-09-17T13:00:20.450187867+05:30"},{"datapoint":9300,"normalizedTime":1568705513,"timestamp":"2019-09-17T13:00:20.450199331+05:30"},{"datapoint":9400,"normalizedTime":1568705514,"timestamp":"2019-09-17T13:00:20.450210255+05:30"},{"datapoint":9500,"normalizedTime":1568705515,"timestamp":"2019-09-17T13:00:20.450218366+05:30"},{"datapoint":9600,"normalizedTime":1568705516,"timestamp":"2019-09-17T13:00:20.450229149+05:30"},{"datapoint":9700,"normalizedTime":1568705517,"timestamp":"2019-09-17T13:00:20.450237776+05:30"},{"datapoint":9800,"normalizedTime":1568705518,"timestamp":"2019-09-17T13:00:20.450245113+05:30"},{"datapoint":9900,"normalizedTime":1568705519,"timestamp":"2019-09-17T13:00:20.450252257+05:30"}]
```
Harkishen-Singh: This should not be changed. Kindly add the previous value. Thanks

tsdb/tsdb.go
```diff
@@ -154,3 +155,23 @@ func (c Chain) Save() bool {
...
 	return true
 }
+
+//GetPositionalPointerNormalized Returns block by searching the chain for the NormalizedTime
+func (c Chain) GetPositionalPointerNormalized(n int64) *Block {
+	c.lengthElements = len(c.chain)
+	jumpSize := int(math.Floor(math.Sqrt(float64(c.lengthElements))))
+	index := jumpSize - 1
+	for n > c.chain[index].NormalizedTime {
+		index += jumpSize
```
Harkishen-Singh: What if the `n` value is not present in the array. Suppose a case where the `n` value is given much larger than the possible time in `time.UnixNano()`, in that case it would lead to an infinite loop and ultimately towards a core dump which is fatal. Pls add a check in this case and report an error and also have a test case checking this whether the code throws error or not when the value is much larger than the possible `time.UnixNano()`.
Thanks!

test-files/loadFromStorage_testdata/test1.json
ankitjena: I changed the normalizedTime, which was same for all the blocks. Traversal wouldn't work other wise.

tsdb/tsdb.go
ankitjena: Doing that

test-files/loadFromStorage_testdata/test1.json
Harkishen-Singh: okay sure

ankitjena:
The build's still failing due to the force push.

tsdb/tsdb.go
```diff
@@ -154,3 +155,40 @@ func (c Chain) Save() bool {
...
+	}
+
+	for n > c.chain[index].NormalizedTime && index < c.lengthElements-jumpSize {
+		index += jumpSize
+	}
+
+	// for c.chain[index].NormalizedTime != n {
+	// 	if c.chain[index].NormalizedTime > n {
+	// 		index--
+	// 	} else {
```
Harkishen-Singh: Please remove these comments. @ankitjena

ankitjena: @Harkishen-Singh done

Issue #41, opened 2019-10-05
Simplify handlers.go


Harkishen-Singh:
The `handlers.go` is too long and confusing since the independent go-routines funcs are written in the same file. This can be simplified by taking independent go-routine functions to different appropriate files, thus reducing the length and complexity of `handlers,go`

ankitjena:
Closing this

PR #48, opened 2019-10-06
Simplify handlers.go #41

flyznex:
- Separate case "start" to new file.
- Use `map[string]string` instead `[]string` to store urlStack. Reduce for loop


Harkishen-Singh:
@flyznex thanks for showing your interest for Bench-routes. Will review asap!

ankitjena:
@flyznex Looks good

src/lib/handle_ping_start.go
```diff
@@ -0,0 +1,86 @@
...
+		Configuration.Config.UtilsConf.ServicesSignal.Ping = "active"
+		_, e := Configuration.Write()
+		if e != nil {
+			panic(e)
+		}
+		go func() {
+			pingConfig := Configuration.Config.Routes
+			pingInterval := getInterval(Configuration.Config.Interval, "ping")
+			if pingInterval == (testInterval{}) {
+				log.Fatalf("interval not found in configuration file for ping")
```
Harkishen-Singh: It would be really good if the function which runs as a different go-routine _(the one selected)_ is kept in this file along with `doPing()` since the newer go-routine is that part in `handlers.go` which increases the complexity.

I would suggest making a newer directory named `handlers`, and keep this file in it. which would contain the code belonging to the function running as an independent go-routine.

src/lib/handle_ping_start.go
```diff
@@ -0,0 +1,86 @@
...
+				return
+			}
+
+			urlStack := make(map[string]string)
+			urlHashMap := make(map[string]string)
```
Harkishen-Singh: thanks

src/lib/handlers.go
```diff
@@ -38,77 +34,7 @@ func HandlerPingGeneral(signal string) bool {
...
 	switch signal {
 	case "start":
-		if pingServiceState == "passive" {
-
-			Configuration.Config.UtilsConf.ServicesSignal.Ping = "active"
-			_, e := Configuration.Write()
-			if e != nil {
-				panic(e)
-			}
-			go func() {
```
Harkishen-Singh: keep this part of code in the current file and transfer the rest in `handlers/handle_ping_start.go` file with `handlers` as a new directory.

src/lib/handle_ping_start.go
aquibbaig: @Harkishen-Singh I would most likely require this logic for having unique routes universal. For example, we need unique routes for every module which have different inputs in each case(say `route url` for ping module, `route url and method` for req-res delay). So, the filtering must be done on the route itself(not the url) which means store the routes in some Data Structure and filter based on route.URL. Thoughts?

aquibbaig: Basically, the idea is to make a routeStack not a urlStack and append/delete based on route.URL. Why decompose the route struct as per URL which seems to be a `variant` property in comparision to the route struct, which is `invariant`

Harkishen-Singh: I understand your point @aquibbaig. `urlStack` is used here since ping is concerned with unique IP addresses than unique routes. however, for most modules, we can also maintain a global `routeStack` as you mentioned.

However, pls note that the `routeStack` will contain IP addresses too, due to the case of different VM instances having different routes required to be tested as well.

Harkishen-Singh: This can be done in a different PR as this PR is specific to **simplification**. if possible, pls try having the above idea in req-res PR.

Thanks!

PR #58, opened 2019-10-07
Req res integration with sockets

### LINKED ISSUE #38: Integrate modules with sockets and tsdb
We need to integrate the modules with the corresponding socket message and tsdb as an end-to-end working model.

aquibbaig:
Info:
- Follows #38


src/lib/modules/response/resp_delay_test.go
```diff
@@ -1,57 +1,49 @@
...
 
-const (
-	threshold = 0
-)
-
-var routes = []Route{
-	Route{
-		"https://www.zairza.in/",
-		"GET",
+var routes = []utils.Routes{
```
golangcibot: `routes` is unused (from `varcheck`)

storage/ping/chunk_ping_bing.com.json
```diff
@@ -0,0 +1 @@
+[]
```
Harkishen-Singh: kindly restore all **ping chunks** from the original code. We don't wanna delete these in each PR, they need to be present at the time of writing tests!

aquibbaig: They are not in the original code

Issue #46, opened 2019-10-05
Make bench-routes forced to undergo Safe exit when Ctrl+C


Harkishen-Singh:
This is important. During the execution of modules involving thousands of goroutines, if the application is forced closed via `Ctrl+C`, there are high chances of corrupted data records that could gather or go un-committed in the TSDB. 

We need a mechanism to make all those unsafe methods undergo a definite procedure whenever a `Ctrl+C` or `SIGINT` signal is triggered.

Reference [link](https://stackoverflow.com/a/11269077)

ganeshpatro321:
Will take up this.

ganeshpatro321:
Is this solved @Harkishen-Singh

Harkishen-Singh:
No. This is still to be covered.

ganeshpatro321:
@Harkishen-Singh I was thinking of checking the service state of all the modules from the config file, see whether the state of the modules are active or passive, if active then stop the service, and then os.exit(1) the app. What do you think about this approach?

Harkishen-Singh:
Yes, moreover, explicitly declaring os.exit(0) should be the answer. Putting 1 will show a failure or error.

ganeshpatro321:
Okay, noted.

Harkishen-Singh:
I think we have a level of this currently. Closing this. We can later re-open when this becomes a priority, probably when we handle a million go-routines.

PR #59, opened 2019-10-07
Add context to graceful shutdown concurrency #46

flyznex:
Add context to graceful shutdown concurrency #46


src/service/sockets.go
```diff
@@ -22,52 +25,70 @@ func home(w http.ResponseWriter, r *http.Request) {
...
 	http.HandleFunc("/", home)
-	http.HandleFunc("/websocket", sockets)
-
-	log.Fatal(http.ListenAndServe(port, nil))
+	http.HandleFunc("/websocket", sockets(ctx))
+	sr := &http.Server{
+		Addr: s.Port,
+	}
+	g.Go(func() error {
+		select {
```
golangcibot: S1000: should use a simple channel send/receive instead of `select` with a single case (from `gosimple`)

src/main.go
```diff
@@ -1,20 +1,45 @@
...
-	service.Runner(port)
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, os.Interrupt, syscall.SIGINT, syscall.SIGTERM)
+	defer func() {
+		signal.Stop(c)
+		cancelFunc()
+	}()
 
+	go func() {
+		select {
```
golangcibot: S1000: should use a simple channel send/receive instead of `select` with a single case (from `gosimple`)

src/lib/modules/ping/ping_test.go
```diff
@@ -65,9 +68,15 @@ func TestHandlerPing(t *testing.T) {
 	initTest()
 	var wg sync.WaitGroup
 	wg.Add(len(urls))
+	ctx, cancelFunc := context.WithCancel(context.Background())
 
 	for _, inst := range urls {
-		go HandlePing(tsdb.GlobalPingChain, &inst, 10, "", &wg, true)
+		go HandlePing(ctx, tsdb.GlobalPingChain, &inst, 10, "", &wg, true)
+	}
+	select {
```
golangcibot: S1000: should use a simple channel send/receive instead of `select` with a single case (from `gosimple`)

Harkishen-Singh:
@flyznex thanks for showing yr interest towards this issue. We actually aimed to safeguard things from a **manual** force close, or in simple terms, prevent forced the closing of application whenever the user signals a  which happens by `Ctrl+C` (SIGINT). But the code submitted in this does not seem to do so. 

This approach would actually stop the existing child processes. This could be helpful only if it is extended to the operations of tsdb as well, since committing the latest data records to the storage servers as a critical operation which might get corrupted in force close.

Moreover, we need to decide on which parent process we require a `cancel` signal. Only then, this approach will be useful. Please try to bring an approach for including cancellation signals at appropriate points.

Thanks!

flyznex:
@Harkishen-Singh, Does your mean is not allow to force close application with `Ctrl+C` (SIGINT)?

Harkishen-Singh:
@flyznex exactly. We need to wait till all the running operations end and the necessary changes are done in the DB, and only then close the application.

Issue #65, opened 2019-10-08
If hostname for a particular route is not correct, entire application crashes.


Pratik32:
**Describe the bug**
If we specify wrong hostname in the config file, the application panic while pinging the incorrect hostname.

**To Reproduce**
Steps to reproduce the behaviour:
1.Replace anyone of the hostname in config file with garbage value.
   e.g I replaced www.google.com to www.gougle.com
2.Test the application with socket-tester
3. In the logs it is evident that, when it tries to ping the incorrect hostname error caught and panic() is called for the given routine.  

**Expected behaviour**
1. Since this is happening for a particular hostname, It should continue monitoring other routes/hostnames (which are correct). 
2. If we call panic in this scenario, it panics the entire application.
3. Expected behavior : For the erroneous hostname, a log message should be printed, and nothing should be added to tsdb.

**Desktop (please complete the following information):**
 - OS: Linux
 - Version : master

Harkishen-Singh:
Thanks for pointing out. We are in the early stages for the development. Would be having similar bugs in other modules. 
Thanks for your interest in the project. Expecting more participation!

muskankhedia:
@Pratik32 Thanks for your interest in the project. On examining the cases, I came to know that on invalid URLs the server doesn't panic instead its redirecting to the IP `154.92.19.168` and the complete process works as expected. So there won't be any requirement of any extra conditions foor Invalid URLs.

Screenshot of the redirection of URL in an invalid input:
![WhatsApp Image 2019-10-09 at 7 09 42 PM](https://user-images.githubusercontent.com/36980003/66506114-5624a600-eaea-11e9-84c7-4edf12818913.jpeg)

P.S. The redirecting behaviour is not observed in all the systems, I will surely figure out the reason soon for the same.

ankitjena:
@muskankhedia that doesn't happen in my system
```
ping: www.gougle.com: Name or service not known

```

muskankhedia:
> @muskankhedia that doesn't happen in my system
> 
> ```
> ping: www.gougle.com: Name or service not known
> ```

@ankitjena Yaa same happening in my system, but working fine with Harkishen. So we need to find the basic reason for redirection to the default URL in some of the system.

Pratik32:
Hi @muskankhedia , thanks. did you try the repro steps that I mentioned in the bug report. Can you check the result with those steps.
[...]

PR #66, opened 2019-10-08
Do not panic the ping go routine if an error is returned from ping command.

Pratik32:
This code change is for fixing bug #65 that I reported.
To fix this, I am Returning an empty string to indicate error. which will be handled in HandlePing() and nothing is added in tsdb for error case.


Pratik32:
There is one cleaner way of catching/indicating error -> Pass 2 channels to CLIPing() one result channel and one error channel. if error occurs, we can send error code through error channel and check it in HandlePing(). Let me know your thoughts on this.

Harkishen-Singh:
Thanks for your interest in the project. Will review asap.

muskankhedia:
@Pratik32 Please have a look at [https://github.com/zairza-cetb/bench-routes/issues/65#issuecomment-540111584](https://github.com/zairza-cetb/bench-routes/issues/65#issuecomment-540111584) for more details regarding the PR.

Please follow the DCO rules while making commits to get the DCO tests passed. For more information about DCO [click here](https://probot.github.io/apps/dco/).

src/lib/modules/ping/ping.go
```diff
@@ -21,6 +21,10 @@ func HandlePing(globalChain []*tsdb.ChainPing, urlRaw *string, packets int, tsdb
 	// launch a goroutine to handle ping operations
 	go utils.CLIPing(urlRaw, packets, chnl)
 	resp := <-chnl
+	if *resp == "" {
```
Harkishen-Singh: if the `ping` works abnormally, then it should return `.... Name or service not known` rather than an empty string, since the golang subprocess module jumps to the next line of execution-only after finishing a particular instruction, and since the mechanism is not normal, it should return an exception than an empty string.
Please have a look.

Harkishen-Singh: Sure. That would be good

Pratik32:
updated the PR. Followed the DCO rule for the new PR. the old commit is till without 'signed-off-by' :)

src/lib/modules/jitter/jitter.go
```diff
@@ -9,7 +9,16 @@ import (
...
+	errChnl := make(chan error)
+	go utils.CLIPing(url, packets, chnl, errChnl)
+	errCode <- errChnl
 	resp := <-chnl
+	if errCode != nil {
+		log.Println(*resp)
+		close(errChnl)
+		close(chnl)
+		*jitter = 0
+		return
```
Harkishen-Singh: could you make a second return param as an error, which would be returned in case of abnormal behaviour. This is becoz setting jitter to 0 will indicate an ideal system than any error. 
We can return the error and log it rather than panic. That should be the case.

What do you think?

src/lib/modules/ping/ping.go
```diff
@@ -16,11 +17,21 @@ const (
...
-	go utils.CLIPing(urlRaw, packets, chnl)
+	go utils.CLIPing(urlRaw, packets, chnl, errChnl)
+	errCode := <-errChnl
 	resp := <-chnl
+	if errCode != nil {
+		log.Println(*resp)
+		close(errChnl)
+		close(chnl)
+		wg.Done()
+		return
```
Harkishen-Singh: same here

Harkishen-Singh:
@Pratik32 you could discuss [here](https://gitter.im/bench-routes/community). That would be really appropriate

src/lib/modules/jitter/jitter.go
Pratik32: Yes. You are right. The error should be propagated back to caller function. I will make that change.
I thought of doing this in the same PR, but I wanted to hear your comments first. will make the change.

Harkishen-Singh:
@Pratik32 please pull latest commits and resolve the conflicts.

PR #70, opened 2019-10-09
create path in storage if not found

### LINKED ISSUE #69: Save to TSDB fails when folder doesn't exist
**Describe the bug**
During init or any test, if folder doesn't exist inside storage, then it throws
```
open: no such file or directory
```

**To Reproduce**
Steps to reproduce the behaviour:
1. Remove any folder from the storage, like `ping` folder
2. Start bench-routes, `make run`
3. It will throw error

**Expected behaviour**
If the folder doesn't exist, it should create one

**Desktop :**
 - OS: ubuntu

ankitjena:
Signed-off-by: ankitjena <ankitjena13@gmail.com>

Fixes #69  .Check for path when writing to HDD, in case path doesn't exist create one.

### DIFF
--- a/tsdb/parse.go
+++ b/tsdb/parse.go
@@ -53,7 +53,10 @@ func checkAndCreatePath(path string) {
 	path = strings.Join(array, "/")
 	_, err := os.Stat(path)
 	if err != nil {
-		os.MkdirAll(path, os.ModePerm)
+		e := os.MkdirAll(path, os.ModePerm)
+		if e != nil {
+			panic(e)
+		}
 	}
 }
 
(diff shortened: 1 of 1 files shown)


tsdb/parse.go
```diff
@@ -45,7 +47,18 @@ func loadFromStorageFloodPing(raw *string) *[]BlockFloodPingJSON {
 	return &inst
 }
 
+func checkAndCreatePath(path string) {
+	array := strings.Split(path, "/")
+	array = array[:len(array)-1]
+	path = strings.Join(array, "/")
+	_, err := os.Stat(path)
+	if err != nil {
+		os.MkdirAll(path, os.ModePerm)
```
golangcibot: Error return value of `os.MkdirAll` is not checked (from `errcheck`)

Issue #54, opened 2019-10-06
Displaying the log message if the server gets disconnected


muskankhedia:
OS: Fedora

#### Describe
There are chances where the network may get disconnected in the mid of the execution of a function. So displaying the log message when the network gets disconnected will aware the user to reconnect the network and the log will be maintained.

Pratik32:
hello @muskankhedia I would like to work on this one. Is there anyone already working on it ?

muskankhedia:
@Pratik32 Thanks for showing your interest in the project. Currently, this issue is open for contributors. You can work on it and feel free to ask any queries.

Pratik32:
Hello, In order to implement this I was thinking of following:
1) the linux's ping utility return 'could not find host < host_name>' if it cannot connect to the given host.(possibly because of no network connection or something wrong with host itself). If ping returns this message we can log the message.(to make user aware of it).
OR
2) In order to check network connectivity we can ping google's DNS : 8.8.8.8.

Pratik32:
Okay, I just realized that there already a method VerifyConnection() in source code to check external network connectivity, I think we can use it too ?

muskankhedia:
> Hello, In order to implement this I was thinking of following:
> 
> 1. the linux's ping utility return 'could not find host < host_name>' if it cannot connect to the given host.(possibly because of no network connection or something wrong with host itself). If ping returns this message we can log the message.(to make user aware of it).
 
Yaa something similar needs to be done as we need to check if the network gets disconnected in the mid of any operation. A condition should be made to check internet connectivity at every instant and aware the user as soon as the network is disconnected.

muskankhedia:
> Okay, I just realized that there already a method VerifyConnection() in source code to check external network connectivity, I think we can use it too ?

VerifyConnection() checks for internet connectivity only once when the server is started. Now we need to implement an all-time internet connectivity check and warn the user as soon as the network gets disconnected at any instant when the server is running.

Pratik32:
@muskankhedia thanks muskan, I have thought of some code change regarding this. will raise a PR once done.

PR #79, opened 2019-10-12
Added code to check externel network connection at specific interval. 

Pratik32:
This w.r.t to #54.

Testing : Tested the code by switching on/off the internet connection.it works as expected.

### DIFF
--- a/src/lib/handlers/handle_ping_start.go
+++ b/src/lib/handlers/handle_ping_start.go
@@ -45,15 +45,14 @@ func doPing(config parser.YAMLBenchRoutesType, urlStack map[string]string, pingI
 			err, _ := utils.VerifyConnection()
 			if !err {
 				log.Println("Not able to connect to externel network please check you internet connection")
-				goto sleep
+			} else {
+				var wg sync.WaitGroup
+				wg.Add(len(urlStack))
+				for _, u := range urlStack {
+					go ping.HandlePing(tsdb.GlobalPingChain, u, 10, u, &wg, false)
+				}
+				wg.Wait()
 			}
-			var wg sync.WaitGroup
-			wg.Add(len(urlStack))
-			for _, u := range urlStack {
-				go ping.HandlePing(tsdb.GlobalPingChain, u, 10, u, &wg, false)
-			}
-
-			wg.Wait()
 		case "passive":
 			// terminate the goroutine
 			log.Printf("terminating ping goroutine\n")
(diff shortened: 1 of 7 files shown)


muskankhedia:
@Pratik32 Thanks for your interest and contribution. 

Some things to keep in mind while creating PRs:-
1. Please use different branches while creating PRs, that will make it manageable and easy to know the requirements and changes for the particular feature/bug and will also help to keep the code clean.
2. Add a GIF or link to the video of the working of the code. That will provide assurity of the working of the code in your system and will decrease conversations (You can install [screencastify](https://chrome.google.com/webstore/detail/screencastify-screen-vide/mmeijimgabbpbgpdklnllpncmdofkcpn?hl=en) extension for screen recording and add the video link in the description).

muskankhedia:
@Pratik32 Great Work!!
Everything is working fine with my system but the log message isn't displayed when the internet is disconnected. Have a look at that. And implement the same for the Jitter and Flood-Ping.

Pratik32:
@muskankhedia , Thanks. Will make sure I include testing results in PRs. Regarding log messages not getting displayed, can you check if your current config is 'active'  because we check for connectivity only in active mode.

src/lib/handlers/handle_ping_start.go
```diff
@@ -42,6 +42,11 @@ func doPing(config parser.YAMLBenchRoutesType, urlStack map[string]string, pingI
 
 		switch config.Config.UtilsConf.ServicesSignal.Ping {
 		case "active":
+			err, _ := utils.VerifyConnection()
+			if !err {
+				log.Println("Not able to connect to externel network please check you internet connection")
+				goto sleep
```
Harkishen-Singh: could we avoid `goto` and just put the remaining part before the next case inside the `else`. That would be really good since using `goto` will cause readability issues once the code becomes large.

src/lib/utils/connection.go
```diff
@@ -9,7 +9,7 @@ import (
 func VerifyConnection() (bool, int) {
 	res, err := http.Get("http://clients3.google.com/generate_204")
 	if err != nil {
-		log.Fatalf("unable to ping external network\n")
+		log.Println("unable to ping external network")
```
Harkishen-Singh: pls use  `log.Printf()` always instead of `log.Println()`. Please change the same in other lines as well.

muskankhedia:
@Pratik32, I checked it well, the config is `active` and also the ping continues as soon as the network is connected. But due to some reasons log message after internet is disconnected is not displayed.

Pratik32:
ss1
Attaching some testing results. (still learning to use screencastify). When the network is disconnected, the messages are getting logged.

Harkishen-Singh:
Looks clean and should work.
Thanks @Pratik32

Pratik32:
Thanks for the review.

Issue #37, opened 2019-10-04
Support for saving log to secondary storage 


Harkishen-Singh:
Currently, we don't support saving of system logs generated by BR. This might be crucial from the debugging point. 

Things to note:
1. Keep this log in some other folder which should be excluded from `git`
2. The secondary storage should capture each and every log generated.

PR #83, opened 2019-10-14
code for saving logs to files.

Pratik32:
Added code to save logs generated by bench-route to secondary storage.
The log file is created in current user's home directory with date & time.
Fixes: #37 

Signed-off-by: Pratik Shinde <pratikshinde320@gmail.com>

This creates a file with name : bench-route-<current-date-time>.log in current user's home directory.
logger writes to both stdout & logfile. We can add a command line option to bench-route utility to enable/disable stdout output.

Testing:
logger
logger2

### DIFF
--- a/src/main.go
+++ b/src/main.go
@@ -1,15 +1,19 @@
 package main
 
 import (
+	"fmt"
 	"github.com/gorilla/websocket"
 	"github.com/zairza-cetb/bench-routes/src/lib/filters"
 	"github.com/zairza-cetb/bench-routes/src/lib/utils"
 	"github.com/zairza-cetb/bench-routes/src/lib/utils/parser"
 	"github.com/zairza-cetb/bench-routes/tsdb"
+	"io"
 	"log"
 	"net/http"
 	"os"
+	"os/user"
 	"strconv"
+	"time"
 )
 
 var (
@@ -22,10 +26,12 @@ var (
 	configuration parser.YAMLBenchRoutesType
 )
 
-func init() {
+const (
+	logFilePrefix = "bench-route-"
+)
 
-	log.SetPrefix("LOG: ")
-	log.SetFlags(log.Ldate | log.Lmicroseconds | log.LstdFlags | log.Lshortfile)
+func init() {
+	setupLogger()
 	log.Printf("initializing bench-routes ...")
 
 	// load configuration file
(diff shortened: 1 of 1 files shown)


Pratik32:
@Harkishen-Singh , can you check why GolangCI is running into error. I had run it locally. no errors observed.

Harkishen-Singh:
@Pratik32 **golang-ci** has been broken after pushing vendor directory without the mods. Would be fixed in following commits. Till then, let's ignore it.

will review asap. Thanks!

src/main.go
```diff
@@ -221,3 +227,24 @@ func main() {
...
+func setupLogger() {
+	currTime := time.Now()
+	currFileName := fmt.Sprint(logFilePrefix, currTime.Format("2006-01-02#15:04:05"), ".log")
+	user, err := user.Current()
+	if err != nil {
+		fmt.Printf("cannot access current user data\n")
+		return
+	}
+	homePath := user.HomeDir
+	logFilePath := homePath + "/" + currFileName
```
Harkishen-Singh: Please store the logs inside the `storage/logs/` _(create one by code if not found)_ directory in bench-routes. We do not want to overcrowd the `/home`.

src/main.go
```diff
@@ -221,3 +227,24 @@ func main() {
...
+	file, err := os.OpenFile(logFilePath, os.O_RDWR|os.O_CREATE|os.O_APPEND, 0666)
+	if err != nil {
+		fmt.Printf("error opening log file : %s\n", logFilePath)
+		return
+	}
+	writer := io.MultiWriter(os.Stdout, file)
+	log.SetOutput(writer)
+	log.SetPrefix("LOG: ")
+	log.SetFlags(log.Ldate | log.Lmicroseconds | log.Llongfile)
```
Harkishen-Singh: please use `log.Lshortfile`. long files log too much to find out the necessary incident.

src/main.go
```diff
@@ -221,3 +227,24 @@ func main() {
...
+	currTime := time.Now()
+	currFileName := fmt.Sprint(logFilePrefix, currTime.Format("2006-01-02#15:04:05"), ".log")
+	user, err := user.Current()
+	if err != nil {
+		fmt.Printf("cannot access current user data\n")
+		return
+	}
+	homePath := user.HomeDir
+	logFilePath := homePath + "/" + currFileName
+	file, err := os.OpenFile(logFilePath, os.O_RDWR|os.O_CREATE|os.O_APPEND, 0666)
```
Harkishen-Singh: please make sure that when the log file is created or closed, the default access should only be **Read**, i.e., write-protected. this is for security concerns as only the owner or admin retains the right to alert the logs, thereby lessening the confusion at the later stages of the product.

src/main.go
Pratik32: @Harkishen-Singh In future if bench-routes is shipped as a standalone binary (without source) then the path has to be changed. hence I chose current user's home for dumping logs. what do you think ?

Harkishen-Singh: @Pratik32 we are yet to plan in that part. For now, we can make a new folder in the `home` as **br-logs** and keep all the logs in it in. That would be better.

Pratik32: Okay. will modify code accordingly.

src/main.go
Pratik32: Done.

src/main.go
Pratik32: Done.

src/main.go
```diff
@@ -236,15 +237,22 @@ func setupLogger() {
...
-	logFilePath := homePath + "/" + currFileName
-	file, err := os.OpenFile(logFilePath, os.O_RDWR|os.O_CREATE|os.O_APPEND, 0666)
+	logDirectoryPath := homePath + "/" + logDirectory
+	err = os.MkdirAll(logDirectoryPath, 0755)
+	if err != nil {
+		fmt.Printf("error creating log directory : %s\n", logDirectoryPath)
+		return
+	}
+	logFilePath := logDirectoryPath + "/" + currFileName
+	file, err := os.OpenFile(logFilePath, os.O_RDWR|os.O_CREATE|os.O_APPEND, 0644)
```
Harkishen-Singh: change `0644` to `0444` to make the generated log **Read-only**

Pratik32: Done.

Issue #84, opened 2019-10-15
Fix GolangCI


Harkishen-Singh:
**Describe the bug**
Golang CI check bot has been failing since the merge of **vendor** directory.

**To Reproduce**
Steps to reproduce the behaviour:
1. On submitting any PR

Harkishen-Singh:
A brief examining of the Golang-CI bot logs concluded that this **might** have occurred due to the absence of go.mod and go.sum files.

Issue #81, opened 2019-10-13
Move to Docker


aquibbaig:
Let's have a container for our application which will make it easier for people(Dev) to pull the application rather than going through all the installation process.

Tjzabel:
I submitted PR #94 to add docker support :+1:

Harkishen-Singh:
@Tjzabel thanks for the contribution. Closing this issue. Thanks.

PR #94, opened 2019-10-25
Add support for Go modules, docker

Tjzabel:
## What This PR Does
* Adds docker support
* Moves from Go vendor to Go modules

Since I updated the INSTALL.md instructions, feel free to let me know if I changed too much, or should remove/add additional details.

### Issues potentially fixed
This PR also potentially fixes issues:
* #84 
* #81 
* #54 (should already be able to be closed)


Tjzabel:
@aquibbaig I tested the functionality of `make run`, and threw up a webpage at `localhost:9090` to make sure I was able to get responses :+1:

INSTALL.md
```diff
@@ -1,8 +1,16 @@
...
+
+1. Make sure [docker](https://www.docker.com/) is installed.
+2. Run `docker build -t bench-routes .`
+3. Run `docker run -p 9090:9090 -it bench-routes`
+
+### Local machine
+
 1. Install Golang +1.10.x (assuming `git` is already installed).
 2. Make a default repository for cloing the project. This should be strictly inside the `GOPATH`. Paste this instruction in your terminal to get started.
 `mkdir -p $GOPATH/src/github.com/zairza-cetb/`.
```
Harkishen-Singh: could you please update the installation instructions with `go mod vendor` so that the users could get the dependencies in the vendor folder.

INSTALL.md
```diff
@@ -1,8 +1,16 @@
...
 
 Please follow the following steps in order to set up your development
 environment.
 
+### Docker
+
+1. Make sure [docker](https://www.docker.com/) is installed.
+2. Run `docker build -t bench-routes .`
+3. Run `docker run -p 9090:9090 -it bench-routes`
+
```
Harkishen-Singh: please mention `superuser permission might be required`

Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM fedora:30
+
+RUN dnf -y update && \
+    dnf -y install make go
+
+COPY . ./app
+
+WORKDIR ./app
+
```
Harkishen-Singh: please implement go testing before starting the application

INSTALL.md
Tjzabel: @Harkishen-Singh Hmm, using the go modules that I added replaces the `vendor/` directory, and thus is no longer necessary.

Dockerfile
Tjzabel: Testing the application in the Dockefile isn't the proper way to do this.

Generally, the developer would:
1. Write new functionality
2. Run `go test` as they develop the new unit tests to make sure everything passes
3. Use docker as the deployment method after everything is working as expected

Otherwise, a second `Dockerfile.dev` can be made to do all of this if you want the development to be fully containerized as well. This new Dockerfile would then spin up the application and do all the testing you want, so the developer doesn't need to install any local dependencies.

INSTALL.md
Tjzabel: I've updated to specify the user needs to be in the docker group in order to run these commands.

Dockerfile
Harkishen-Singh: The reason why we want test just before the run command in docker is just to ensure the user that the build he is using for deployment has passed all the unit-tests in order to ensure that no preliminary errors exist in the deployment build. 

Note: we dont aim for testing in docker. This is just to ensure that the latest version he is building in docker is free from all basic probs which have been covered in the unit-testing.

PR #95, opened 2019-10-26
Used binary search for finding block

### LINKED ISSUE #93: Why can't we use Binary Search rather than using Jump search?
In TSDB, the blocks are searched using JUMP search.
But binary search will have a less number of steps to be executed, and would be a faster alternative to Jump search.

biswajitpatra:
Want to resolve #93

### DIFF
--- a/tsdb/tsdb.go
+++ b/tsdb/tsdb.go
@@ -3,7 +3,6 @@ package tsdb
 import (
 	"errors"
 	"log"
-	"math"
 	"sync"
 	"time"
 	"unsafe"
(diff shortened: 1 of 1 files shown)


tsdb/tsdb.go
```diff
@@ -506,90 +505,79 @@ func (c *Chain) GetPositionalPointerNormalized(n int64) (*Block, error) {
...
 	c.LengthElements = len(c.Chain)
-	jumpSize := int(math.Floor(math.Sqrt(float64(c.LengthElements))))
-	index := jumpSize - 1
 
 	if c.Chain[c.LengthElements-1].NormalizedTime < n || c.Chain[0].NormalizedTime > n {
 		return nil, errors.New("Normalized time not in Chain range")
 	}
+	l, r, m := 0, c.LengthElements, 0
 
-	for n > c.Chain[index].NormalizedTime && index < c.LengthElements-jumpSize {
```
Harkishen-Singh: how are you going to check the underflow and overflow condition of the query? I find this removed from yr code

tsdb/tsdb.go
```diff
@@ -506,90 +505,79 @@ func (c *Chain) GetPositionalPointerNormalized(n int64) (*Block, error) {
...
-	}
-
-	for c.Chain[index].NormalizedTime < n {
-		index++
-	}
+	for l <= r {
+		m = l + (r-l)/2
+		if c.Chain[m].NormalizedTime == n {
+			return &c.Chain[m], nil
+		}
```
Harkishen-Singh: I think the infinite loop is not being stopped in cases of an invalid query, or a block not found. This will lead to infinite loop, thus freezing the app. I am correct?

tsdb/tsdb.go
```diff
@@ -506,90 +505,79 @@ func (c *Chain) GetPositionalPointerNormalized(n int64) (*Block, error) {
...
 	c.LengthElements = len(c.Chain)
-	jumpSize := int(math.Floor(math.Sqrt(float64(c.LengthElements))))
-	index := jumpSize - 1
-
 	if c.Chain[c.LengthElements-1].NormalizedTime < n || c.Chain[0].NormalizedTime > n {
 		return nil, errors.New("Normalized time not in Chain range")
 	}
+	l, r, m := 0, c.LengthElements, 0
 
-	for n > c.Chain[index].NormalizedTime && index < c.LengthElements-jumpSize {
```
Harkishen-Singh: same

tsdb/tsdb.go
biswajitpatra: the loop will terminate due to a condition in the for loop  ,that is ,l<=r

Harkishen-Singh:
Since the normalized time is serial in nature, better check the out of range condition initially and proceed only if within the range.

tsdb/tsdb.go
biswajitpatra: I think there is no need for overflow/underflow condition required for searching an element.

Harkishen-Singh: We want to avoid situations of false searching. This is to take care of logical errors.

biswajitpatra: the above if condition checks for the false searching

Issue #40, opened 2019-10-05
Initial checking of configuration rules


Harkishen-Singh:
We have no checks for verifying the configuration file `local-config.yml`. This should be done just before loading the config file into the bench-routes environment.  In case of mismatch, an error should be thrown, thus terminating the system.

JayjeetAtGithub:
Hi @Harkishen-Singh. I would like to work on this too. What would be the rules to validate the yaml file ?

muskankhedia:
@JayjeetAtGithub Thank you for your interest in the project. There are various ways to check the YAML file. One of them is, Initially, you need to check the pattern followed as it should be same as the present one and all the fields are mandatory, so there should be a specific and valid value for every key in the file. And from this, it can be confirmed that the YAML file is valid, ready for parsing and loaded in the server.

JayjeetAtGithub:
@muskankhedia That makes sense. Thanks. I will try to send a PR on this.

JayjeetAtGithub:
@muskankhedia Are we required to show all the errors in the config file and then fail or we can simply fail on encountering the first error ? Like showing one error at a time.

Harkishen-Singh:
I think, we have this now. Closing.

PR #112, opened 2019-12-01
Issue #40 Add validation rules for config file

JayjeetAtGithub:
Signed-off-by: Jayjeet Chakraborty <jc.github@rediffmail.com>

### DIFF
--- a/src/lib/utils/parser/config-validator.go
+++ b/src/lib/utils/parser/config-validator.go
@@ -1,0 +1,106 @@
+package parser
+
+import (
+	"log"
+)
+
+// ValidateRoutesProp validates the `routes` property
+// in the configuration file.
+func ValidateRoutesProp(routes []Routes) {
+	if len(routes) == 0 {
+		log.Fatalf("`routes` property is missing.")
+	} else {
+		for i, route := range routes {
+			if route.Method == "" {
+				log.Fatalf("`method` property for route #%d is missing.", i+1)
+			}
+			if route.Route == "" {
+				log.Fatalf("`route` property for route #%d is missing.", i+1)
+			}
+			if route.URL == "" {
+				log.Fatalf("`url` property for route #%d is missing.", i+1)
+			}
+			if len(route.Header) != 0 {
+				for j, header := range route.Header {
+					if header.OfType == "" {
+						log.Fatalf("`type` property for header #%d of route `%s` is missing.", j+1, route.Route)
+					}
+					if header.Value == "" {
+						log.Fatalf("`value` property for header #%d of route `%s` is missing.", j+1, route.Route)
+					}
+				}
+			}
+			if len(route.Params) != 0 {
+				for j, param := range route.Params {
+					if param.Name == "" {
+						log.Fatalf("`name` property for param #%d of route `%s` is missing.", j+1, route.Route)
+					}
+					if param.Value == "" {
+						log.Fatalf("`value` property for param #%d of route `%s` is missing.", j+1, route.Route)
+					}
+				}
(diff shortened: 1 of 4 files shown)


JayjeetAtGithub:
@Harkishen-Singh @muskankhedia Not sure why the build is failing. Can you please check ?
Also, Please review my PR. Thanks.

muskankhedia:
@JayjeetAtGithub Nice Work! Some minor changes required, please try to create a new module for checking the `local-config file` as created for `parsing` or `ping` as it will help to maintain the code and different modules for different features or checks will be defined.

JayjeetAtGithub:
Ok ! Will do it.

JayjeetAtGithub:
@muskankhedia Can we just create a `config_validator.go` in the `parser` package ? This way we can keep all the parsing related logic in the same pkg.

muskankhedia:
@JayjeetAtGithub Yaa sure. Sounds better

JayjeetAtGithub:
> @muskankhedia Can we just create a `config_validator.go` in the `parser` package ? This way we can keep all the parsing related logic in the same pkg.

@muskankhedia  I updated the PR with this change. Please review. Thanks.

src/lib/utils/parser/config-validator.go
```diff
@@ -0,0 +1,106 @@
...
+			if len(route.Header) != 0 {
+				for j, header := range route.Header {
+					if header.OfType == "" {
+						log.Fatalf("`type` property for header #%d of route `%s` is missing.", j+1, route.Route)
+					}
+					if header.Value == "" {
+						log.Fatalf("`value` property for header #%d of route `%s` is missing.", j+1, route.Route)
+					}
+				}
+			}
```
muskankhedia: @JayjeetAtGithub For now, I think we don't need to make `header` field mandatory. Can you please change it for now?

src/lib/utils/parser/config-validator.go
```diff
@@ -0,0 +1,106 @@
...
+			if len(route.Params) != 0 {
+				for j, param := range route.Params {
+					if param.Name == "" {
+						log.Fatalf("`name` property for param #%d of route `%s` is missing.", j+1, route.Route)
+					}
+					if param.Value == "" {
+						log.Fatalf("`value` property for param #%d of route `%s` is missing.", j+1, route.Route)
+					}
+				}
+			}
```
muskankhedia: Same goes for this. Please have a look and implement the same.

Harkishen-Singh:
@JayjeetAtGithub any updates?

JayjeetAtGithub:
@Harkishen-Singh Sorry for the gap. I will get back to it tomorrow. Does that work ?

Harkishen-Singh:
@JayjeetAtGithub perfect!

JayjeetAtGithub:
Thanks

src/lib/utils/parser/config-validator.go
JayjeetAtGithub: I just checked and realized that `header` field is not mandatory in the above code. I am checking for validation of headers only if it is present.

src/lib/utils/parser/config-validator.go
JayjeetAtGithub: same for `params`.

JayjeetAtGithub:
@muskankhedia  looks like the requested changes are already implemented. Can you please double check ? Thanks a lot.

muskankhedia:
@JayjeetAtGithub Lgtm, @Harkishen-Singh Please have a look.

PR #113, opened 2019-12-02
systemd unit file for bench-routes.

Pratik32:
This adds a systemd unit file for managing bench-routes as service.
The service shall serve following purpose:
1) Start basic monitoring functionality (Configurable)
2) Start listening for incoming requests.
3) When service is stopped (using systemctl stop) the application is shutdown.

**NOTE**: There are few things that needs to be in place before this service can be used :

1) The bench-routes binary needs to be placed in /bin directory after installation
2) Different paths- config file, tsdb directory, logs directory needs to be changed (with current code it expects everything in same directory)
3) A way to gracefully shutdown the bench-route service. (If I am correct this is not done yet ?)

**Testing** : In-order to test this service, I made some changes in my local src code:
1) Accept command arguments for controlling the service from outside.
2) Kept bench-routes binary in /bin
3) Changes different path variables to dump data generated by bench-routes (i.e logs , db, local config file)
The above code change is not included in this commit


Harkishen-Singh:
@aquibbaig @ankitjena @muskankhedia @ganeshpatro321  could you please have a look.

Harkishen-Singh:
@Pratik32 Thank you for taking this up. We are in mid of a time-crunch. Will do soon. Thanks.

bench-routes.service
```diff
@@ -0,0 +1,12 @@
...
+
+[Unit]
+Description= Systemd bench-routes service
+After=network-online.target network.target
+Requires=network-online.target network.target
+
+[Service]
+Type=simple
+RemainAfterExit=yes
+ExecStart=/home/ps/bench-routes 9090 start > /dev/null  2>&1
```
aquibbaig: I haven't followed the discussion yet, but what is this `ps` option here for? Is it a path for your own root directory?

Harkishen-Singh:
@Pratik32 could you please share the resource from where you learnt this. I am unaware of this feature and it would be good learning for me. Also, would be helpful if you can describe the entire process.

bench-routes.service
Pratik32: @aquibbaig  my bad. that's my username. my bench-routes binary is kept there.
'ExecStart' is the absolute path name of the executable that will run when the service is started.
It should be '/bin/bench-routes'.

Pratik32:
a unit is basically a service/daemon/mount points OR in general a "resource" that needs to be brought up when system starts. A unit file is a configuration file that tells systemd process about how to control it.
the basic work flow of bench-route service will be:
1) systemd will start the service by running the bench-routes binary.
     -- This should start the basic monitoring task. (if configured  - yes.)
     -- It will register to a predefined port and starts listening to http/https type requests.
     -- The service will have a direct dependency on network.target/network-online.target (you can read
         about 'targets' in systemd in below links). In simple terms - bench-routes will not be triggered by 
         systemd if machine is not connected to any network interface.
2) Once the service is up and running, it can stopped at will using 'systemctl stop' , This will gracefully
    bring down the entire bench-routes application.
3) It can be started/stopped at will using systemctl interface.

More about systemd units , targets - [LINK](https://www.digitalocean.com/community/tutorials/understanding-systemd-units-and-unit-files)

aquibbaig:
@Pratik32 NIT: While creating a PR, please **add a reference to the issue it will be solving** which makes it easy to see which issues are being worked up (references). This would be useful for beginners looking for first-timer-issues. Thanks!

bench-routes.service
Harkishen-Singh: Please change this to the `bench-routes` since after placing in the `bin`, this can be directly accessed

bench-routes.service
```diff
@@ -0,0 +1,12 @@
...
+ExecStop=/home/ps/bench-routes 9090 stop   > /dev/null  2>&1
```
Harkishen-Singh: Do we really need this?

Harkishen-Singh:
Also, we need to have installation script which should place the `bench-routes.service` file inside the `systemd/system`. I think this can be done via shell being run under sudo. What do you think?

bench-routes.service
Pratik32: thanks for reviewing.
My comments:
1) We can get rid of 'stop' & hence get rid of ExecStop.Because the default behavior of ExecStop is to kill the process. (at the end that's what we want to achieve)
2) The reason why I silenced the command o/p with /dev/null 2>&1 is to make sure we don't flood the systemd logs. 
3) I think we should provide absolute path of the bench-routes binary. I have seen systems that don't include /bin OR /usr/bin in PATH variable. Providing an absolute path will always give the location of binary.

bench-routes.service
Pratik32: Yes ExecStop can be removed.

Pratik32:
Yes. I can write the installation script. and include it as part of this commit. :)

Harkishen-Singh:
@Pratik32 can you please go ahead with the installation script. Then we can merge this. Thanks.

Harkishen-Singh:
Merging for now. At later stages, we can implement this on a full scale. Thanks.

PR #121, opened 2019-12-19
Integrating UI project with the Bench-routes main stream

Harkishen-Singh:
This PR unifies BR development. It presumes all the git history from `bench-routes-electron` project so that the project can be archived and further development can be carried out here.

Please check out the branch [integration-ui-main](https://github.com/zairza-cetb/bench-routes/tree/integration-ui-main) for better understanding.

Reference: [link](https://stackoverflow.com/a/47571452)

**Do-not squash. Squashing might lead to losing important contributors from the `bench-routes-electron` project.**


Harkishen-Singh:
@aquibbaig I couldn't understand that why should we maintain the version. 
Do you mean to Target the next major UI change that should be brought in v2.0 since if that's the case, we have to create a new react app there and completely discard the previous when we switch.

PR #124, opened 2019-12-21
Module signal operations integration in BR-UI

ganeshpatro321:
Signed-off-by: ganeshpatro321 <ganeshpatro321@gmail.com>

Merging of this PR will enable to start module operation (starting and stopping different modules) from the electron app.


dashboard/v1.0/src/components/benchmarks/FloodPing.tsx
```diff
@@ -161,13 +161,42 @@ export default class FloodPing extends Component<
...
+      case 'stop':
+        this.connection.signalFloodPingStop().then(res => {
+          if (res.data) {
+            alert('Flood Ping routine stopped');
+          }
+        });
+        break;
+    }
+  };
+
```
Harkishen-Singh: Can we make this a new independent functional component since this is common for every module. Further, this will involve more than `start`, `stop` in the later stage, so we have to write each functionality 4 times. So, better if we can make this into one component.

local-config.yml
```diff
@@ -51,6 +51,6 @@ utils:
     mean: 0
   services-state:
     ping: passive
-    flood-ping: passive
+    flood-ping: active
```
Harkishen-Singh: please make this passive

src/main.go
```diff
@@ -169,6 +169,7 @@ func main() {
 			// ping
 			case "force-start-ping":
 				// true if success else false
+				fmt.Printf("receiving\n")
```
Harkishen-Singh: please remove this

src/main.go
```diff
@@ -210,6 +211,7 @@ func main() {
 
 				// Get config routes details
 			case "route-details":
+				fmt.Printf("Receiving Route details")
```
Harkishen-Singh: please remove this

local-config.yml
ganeshpatro321: Does this happen automatically, without manual override? Coz, I dont remember changing this.

src/main.go
ganeshpatro321: Shit

dashboard/v1.0/src/components/benchmarks/FloodPing.tsx
ganeshpatro321: Shall I do this now? , Or we can keep this for new contributors as a good first issue?

Harkishen-Singh: better to keep this for the new contributors. Please open a new issue and create a new tag `react-ui` and assign it. Also, give the `good first issue` as well.

local-config.yml
Harkishen-Singh: It happens automatically. Still, please make that to `passive` and stage it.

PR #132, opened 2019-12-25
Facebook URL

Jayashree-panda:
The bing url was showing error with any of these module. So bing is being replaced with facebook.


Harkishen-Singh:
Facebook doesn't work as well. Replace with `youtube.com`.

Harkishen-Singh:
`git pull` should now solve the build fail. @Jayashree-panda

Harkishen-Singh:
I think we should close this. Been more than a month with no activity.

Issue #118, opened 2019-12-07
New brlogger for supporting independent log writes to secondary storage


Harkishen-Singh:
The current logger writes into the terminal (even trivial data which is useless for the terminal thereby spamming the terminal, but for tracing, would be important and hence flushing in the secondary storage) and the log file as well. We need a new logger which can write specific data to the log _(secondary storage)_  and some specific data to terminal and others to both terminal and secondary storage.

Further on later stages, this will implement the compression scheme to compress logs.

ganeshpatro321:
#134

PR #134, opened 2020-01-02
Enhanced log support for better logging.

ganeshpatro321:
This PR restructures the logging in bench-routes. Two different logger are created:
- `TerminalandFileLogger` : This logs to both terminal and secondary storage file
- `FileLogger`: This logs only to the secondary storage file

Logger can be used by importing form `"github.com/zairza-cetb/bench-routes/src/lib/utils/logger"`

There are two functions available for the logger, 
- `logger.Terminal(msg string, code string)` : logs to both terminal and the secondary storage file
- `logger.File(msg string, code string)` : logs only to the secondary storage file.

Codes:
 - `"p"` : for Println()
 - `"f"` : for Fatalln()
 - `"pa"` : for Panicln()


ganeshpatro321:
This solves #118

local-config.yml
```diff
@@ -53,4 +53,4 @@ utils:
     ping: passive
     flood-ping: passive
     jitter: passive
-    req-res-delay-or-monitoring: passive
\ No newline at end of file
+    req-res-delay-or-monitoring: passive
```
Harkishen-Singh: This new line might be a mistake. Please fix this.

src/lib/utils/logger/logger.go
```diff
@@ -0,0 +1,64 @@
...
+	"os/user"
+	"time"
+)
+
+// FileLogger Logs into the secondary storage file
+var FileLogger *log.Logger
+
+// TerminalandFileLogger Logs into the secondary storage file and terminal
+var TerminalandFileLogger *log.Logger
+
```
Harkishen-Singh: i think creating an interface for this should be better.

This should be a better way:

```
type Log struct {
  dir  string  // can be used to initialise the dir for logging. for now lets keep this explicitly as br-logs
  perm string // for permissions of opening or closing the file
}

func New() (*os.File, bool) { // this will do the work of SetupLogger()
  ...
}

func (lg Log) Terminal(message string, code string) { // code will be used to know whether the log is a normal
// message, a warning or a fatal.
  ....
}

func (lg Log) TFile(message string, code string) {
  ...
}

```

This should be the ideal way of creating any module. This will also minimize the long functional names as well and make maintaining easier.

Then, we can use this as,
```
import "github.com/zairza-cetb/..../logger"

log := logger.Log{
  dir: "br-logs",
  perm: "0755",
}

log.Terminal("this is a log", "n")

```

n -> normal
w -> warning
f -> fatal

src/lib/utils/logger/logger.go
```diff
@@ -0,0 +1,64 @@
...
+	generalWriter := io.Writer(file)
+	terminalWriter := io.MultiWriter(os.Stdout, file)
+
+	FileLogger = log.New(generalWriter, "LOG:\t", log.Ldate|log.Lmicroseconds|log.Lshortfile)
+	TerminalandFileLogger = log.New(terminalWriter, "LOG:\t", log.Ldate|log.Lmicroseconds|log.Lshortfile)
+
+}
+
+// SetupLogger sets up files for logging
+func SetupLogger() (fp *os.File, ok bool) {
```
Harkishen-Singh: Please update this according to the above said.

src/lib/utils/logger/logger_test.go
```diff
@@ -0,0 +1,15 @@
...
+)
+
+func TestLogger(t *testing.T) {
+	file, ok := SetupLogger()
+	if !ok {
+		t.Errorf("Setting up the logger failed")
+	} else {
+		t.Logf("Logger file created")
+		t.Logf("%s", file.Name())
+	}
```
Harkishen-Singh: Also, please log some messages to ensure that there is no error happenings.

src/lib/utils/logger/logger.go
ganeshpatro321: Some fatal logs, ask for an extra parameter, we will have to write all the cases,

ganeshpatro321: I thought of the same way, to create a different function and ask for codes for identifying the normal, fatal, warning logs, but it'd be simple if we let the developers use all the functionalities that log provides without having a layer on it to ask for codes.

Check the index here. https://golang.org/pkg/log/

src/lib/utils/logger/logger_test.go
ganeshpatro321: I dint get you here.

src/lib/utils/logger/logger.go
Harkishen-Singh: If they ask for extra parameters, then use the spread operator. The length of the spread operator can be used to differentiate between various types. If that doesnot work, using parameter as an `interface{}` is always an open option.

src/lib/utils/logger/logger_test.go
Harkishen-Singh: Functions involve complex logics, if not now then later. An end-to-end test should cover the output as well. So, basically we use a normal purpose of the function which over here is logging. If the normal operation is being able to perform by the function, then the function is good.  Else, an implicit panic would be automatically returned by some low-level function.

local-config.yml
ganeshpatro321: It was already there I think. Are you sure you want me to remove this? @Harkishen-Singh

src/lib/utils/logger/logger.go
ganeshpatro321: I don't see any problem in keeping it raw, and moreover, are you sure you want to call the .New() function in every package, I think this is gonna create multiple log files. @Harkishen-Singh

Harkishen-Singh: No. We will declare once inside the globalvars file. The reason I am saying is that then we can use this as a completely independent module and then we can publish this and would be useful for other projects or organisations.

ganeshpatro321: I am still having serious doubts regarding the design, specially to integrate all the functions that the raw package provides. Anyways, lets schedule a call whenever you are free, we will come to a conclusion and move forward.

ganeshpatro321:
@Harkishen-Singh Good to merge.

src/lib/modules/jitter/jitter_test.go
```diff
@@ -25,7 +25,7 @@ var (
 )
 
 func initTest() {
-	log.Printf("Testing jitter")
+	logger.Terminal("Testing jitter", "p")
```
Harkishen-Singh: log.Printf() should be here.

src/lib/modules/notification/notify_test.go
```diff
@@ -1,16 +1,17 @@
...
 	"testing"
+
+	"github.com/zairza-cetb/bench-routes/src/lib/utils/logger"
 )
 
 func TestSendNotification(t *testing.T) {
 	dir, err := os.Getwd()
 	if err != nil {
-		log.Fatal(err)
+		logger.Terminal(err.Error(), "f")
```
Harkishen-Singh: logger is not needed in `_test` files. Please use the conventional log.

src/lib/utils/connection.go
```diff
@@ -1,17 +1,18 @@
...
 // VerifyConnection verifies the connectivity of the program with the external network
 func VerifyConnection() (bool, int) {
 	res, err := http.Get("http://clients3.google.com/generate_204")
 	if err != nil {
-		log.Printf("unable to ping external network\n")
+		logger.Terminal("unable to ping external network", "p")
 		return false, -1
 	}
-	log.Printf("external connection verified\n")
+	logger.Terminal("external connection verified", "p")
```
Harkishen-Singh: change this to `logger.File`

src/lib/utils/logger/logger.go
```diff
@@ -0,0 +1,123 @@
...
+	if err != nil {
+		panic(err)
+	}
+
+	generalWriter := io.Writer(file)
+	terminalWriter := io.MultiWriter(os.Stdout, file)
+
+	fileLogger = log.New(generalWriter, "LOG:\t", log.Ldate|log.Lmicroseconds)
+	terminalandFileLogger = log.New(terminalWriter, "LOG:\t", log.Ldate|log.Lmicroseconds)
+
```
Harkishen-Singh: extra line can be removed

src/lib/utils/logger/logger.go
```diff
@@ -0,0 +1,123 @@
...
+	_, file, line, _ := runtime.Caller(depth)
+	location := chopPath(file) + ":" + strconv.Itoa(line)
+	return location
+}
+
+// File logs to the secondary storage.
+// *code*:
+// "p" -> Println()
+// "f" -> Fatalln()
+func File(msg string, code string) {
```
Harkishen-Singh: `// "pa" -> Panicln()`

src/lib/utils/logger/logger.go
```diff
@@ -0,0 +1,123 @@
...
+		logMessage := logLocater(2) + " " + msg
+		fileLogger.Panicln(logMessage)
+	}
+}
+
+// Terminal logs to the secondary storage and the terminal.
+// *code*:
+// "p" -> Println()
+// "f" -> Fatalln()
+func Terminal(msg string, code string) {
```
Harkishen-Singh: `// "pa" -> Panicln()`

src/lib/modules/jitter/jitter_test.go
Harkishen-Singh: since this is a test file.

ganeshpatro321:
@Harkishen-Singh Updated.

PR #135, opened 2020-01-03
req-res-delay monitoring graph 

Jayashree-panda:


src/lib/modules/response/resp_delay.go
```diff
@@ -22,9 +21,11 @@ const (
 // in a Route struct and supplies it to a function in turn to handle it accordingly. We create
 // channels to run tests for each route in parallel, speeding up the process
 func HandleResponseDelayForRoute(responseChains []*tsdb.Chain, route parser.Routes, tsdbNameHash string, wg *sync.WaitGroup) {
-	routeSuffix := filters.RouteDestroyer(route.URL + "_" + route.Route)
+	// routeSuffix := filters.RouteDestroyer(route.URL + "_" + route.Route)
+
 	// Init paths for request-response-monitoring
-	path := PathReqResDelay + "/" + "chunk_req_res_" + routeSuffix + ".json"
+	path := PathReqResDelay + "/" + "chunk_req_res_" + tsdbNameHash[12:len(tsdbNameHash)] + ".json"
```
golangcibot: S1010: should omit second index in slice, s[a:len(s)] is identical to s[a:] (from `gosimple`)

ganeshpatro321: I just did this to make it work, this is not the best way, please restructure it, check how it works with other modules.

src/lib/modules/response/resp_delay.go
```diff
@@ -22,9 +21,11 @@ const (
 // in a Route struct and supplies it to a function in turn to handle it accordingly. We create
 // channels to run tests for each route in parallel, speeding up the process
 func HandleResponseDelayForRoute(responseChains []*tsdb.Chain, route parser.Routes, tsdbNameHash string, wg *sync.WaitGroup) {
-	routeSuffix := filters.RouteDestroyer(route.URL + "_" + route.Route)
+	// routeSuffix := filters.RouteDestroyer(route.URL + "_" + route.Route)
```
ganeshpatro321: clean the code, and change the naming conventions, do not keep it as a hash, if its not a hash value.

Jayashree-panda: @ganeshpatro321 Yeah! I will be modifying the above 2 parts.

ganeshpatro321:
@Jayashree-panda Only push the graph code, and let the golang code be it is, we have to do some serious restructuring for the monitoring part, and the way its creating chunk file, we will keep this point as a discussion for the team.

Jayashree-panda:
@ganeshpatro321 the graphs won't work if we dont change the golang code

ganeshpatro321:
Yeah @Jayashree-panda I know, let the code be, its tested and working, we have to pick a way how we handle the golang code, that I will go through all the modules and we can decide.

Jayashree-panda:
@ganeshpatro321 Yeah! DONE...

Harkishen-Singh:
I couldn't get this. The service is able to read the time-series successfully and returns with complete success to the view via sockets _(check the `data` field)_. Then, why do we need to edit the golang code?

Jayashree-panda:
I have re-written the changed things. Like, I have restored the changed golang stuffs.

ganeshpatro321:
@Harkishen-Singh 

> I couldn't get this. The service is able to read the time-series successfully and returns with complete success to the view via sockets _(check the `data` field)_. Then, why do we need to edit the golang code?

Are you checking with the masters code, or with this PRs?

Harkishen-Singh:
The origin master code which I had done by writing the querier for req-res. That had everything running. It did had response coming to the frontend as well. Then where are we missing the point?

ganeshpatro321:
No it isn't I guess,  the code was querying for a different chunk file name, and the chunk file names created were different, well this is what I saw in jayshree's laptop, and she hadnt changed any code before that.

The bench-routes was panicking by saying, "db not found", so I debugged and found the chunk that was being created are of different name as that of querying. @Harkishen-Singh

Harkishen-Singh:
Well, then in that case, I think the DB would have got corrupted during the process. Let me have a look again and will get back to this soon. However, another attempt can be made by completely deleting the storage folder and starting the machine all over again, since I remember changing the DB read code. Deleting the storage folder should solve it. @ganeshpatro321

ganeshpatro321:
We did that, deleting the entire storage folder, the same thing happened. Anyways, let me check the same thing in my system. @Harkishen-Singh

aquibbaig:
@Jayashree-panda What are you caught up with?

Harkishen-Singh:
Its been a long time and no update on this. Further, this was solved by https://github.com/zairza-cetb/bench-routes/pull/141.

Hence closing this.

PR #140, opened 2020-01-06
Cleaning up, when ctrl^c called.

ganeshpatro321:
Currently, this code just calls the Handle modules functions with stop signal, in case the status of service is active in the config file.

This makes sure that the status of all the modules/services are passive before exiting or closing bench-routes.

@Harkishen-Singh


ganeshpatro321:
@Harkishen-Singh I think we can merge this PR, as this ensures that the config file is restored, when bench-routes stops with ctrl^c.

We can implement the elimination of go routines with select statement in a following PR, as it requires some discussion regarding the design of various things such as:
- Passing of channels through different packages.
- Using of buffered channels for stopping goroutines
- Switch statements is already doing the work of returning from the routine, in 1st layer, we will have to discuss regarding the implementation of select statements. Select statement implementation is good to go with the 2nd layer implementation.

P.S: 1st layer is the goroutine called in general handler, 2nd layer id the goroutine called in handlers such as handle_ping_start, etc.

ganeshpatro321:
@Harkishen-Singh Updates?

src/main.go
```diff
@@ -95,12 +99,51 @@ func init() {
 	logger.Terminal("Bench-routes is up and running", "p")
 }
 
+func cleanup() {
+	logger.Terminal("No. of go routines running are: "+strconv.Itoa(runtime.NumGoroutine()), "p")
```
Harkishen-Singh: in place of `logger.Terminal("No. of go routines running are: "+strconv.Itoa(runtime.NumGoroutine()), "p")`,
```
logger.Terminal(fmt.Sprintf("Alive %d go-routines", strconv.Itoa(runtime.NumGoroutine())), "p")
```

src/main.go
```diff
@@ -95,12 +99,51 @@ func init() {
 	logger.Terminal("Bench-routes is up and running", "p")
 }
 
+func cleanup() {
```
Harkishen-Singh: can we have this function after main()?

ganeshpatro321: ok

Harkishen-Singh:
I think, we still have not achieved with this. Just ran ping and jitter and then ctrl^c, and it stopped ungracefully. Please have a look into this. If possible, pls make a working video.

Screenshot from 2020-01-07 21-21-01

PR #155, opened 2020-02-12
Calculated response length for all requests

muskankhedia:
Currently, the value for the response length for every request has been stored as -1. In this PR, the response length for every request has been calculated.


src/lib/modules/response/resp_delay.go
```diff
@@ -57,8 +58,13 @@ func HandleGetRequest(url string) utils.Response {
...
 	start := time.Now().UnixNano()
 	resp := *utils.SendGETRequest(url)
+	defer resp.Body.Close()
+	content, err := ioutil.ReadAll(resp.Body)
 
-	resLength := resp.ContentLength
+	if err != nil {
+		panic(err)
+	}
+	resLength := int64(len(content))
```
Harkishen-Singh: The default output of `len()` is `int`. So, even if we increase the output range to a 64-bit like here, the output will be the range of 32-bit only, but wasting the extra 32-bit since the extra range is of no use. So, we should rather keep this to `int` type only.

local-config.yml
```diff
@@ -20,7 +20,7 @@ routes:
   - name: q
     value: bench-routes
 - method: GET
-  url: https://www.bing.com
+  url: https://www.facebook.com
```
Harkishen-Singh: I think this was failing in some platforms. Anyways, if we need an extra URL for the test, we can create a new route, rather than changing the existing test URL.

muskankhedia: Yaa sure.. I will update it accordingly.

src/lib/modules/response/resp_delay.go
muskankhedia: The response length stored in the structure type `response`, has a key with `ResLength` of type int64, so I converted the value to type int64. If changing the type of the element will be fine, I will update that.

Harkishen-Singh: Yeah sure. We can change the type of the struct.

PR #158, opened 2020-02-25
Global layout component

aquibbaig:
App.js needs to be singular and free of all nested components, only subscribed to global stores and persist configurations. Common components for each page should come under a global layout component. 

### DIFF
--- a/dashboard/v1.0/src/App.tsx
+++ b/dashboard/v1.0/src/App.tsx
@@ -1,20 +1,13 @@
 import React from 'react';
 import './App.css';
-import Header from './components/layouts/Header';
-import Sidebar from './components/layouts/Sidebar';
+import BaseLayout from './components/layouts/BaseLayout';
 import Navigator from './router/Navigation';
 
 function App() {
   return (
-    <div className="App">
-      <div className="sidebar">
-        <Sidebar />
-      </div>
-      <div className="inner-component">
-        <Header />
-        <Navigator />
-      </div>
-    </div>
+    <BaseLayout>
+      <Navigator />
+    </BaseLayout>
   );
 }
 
--- a/dashboard/v1.0/src/components/layouts/BaseLayout.tsx
+++ b/dashboard/v1.0/src/components/layouts/BaseLayout.tsx
@@ -1,0 +1,17 @@
+import React from 'react';
+import Header from './Header';
+import Sidebar from './Sidebar';
+
+export default function BaseLayout(props) {
+  return (
+    <div className="App">
+      <div className="sidebar">
+        <Sidebar />
+      </div>
+      <div className="inner-component">
+        <Header />
+        {props.children}
+      </div>
+    </div>
+  );
+}


dashboard/v1.0/src/components/layouts/GlobalLayout.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import Header from './Header';
+import Sidebar from './Sidebar';
+
+export default props => {
```
ganeshpatro321: props keyword here can kind of mislead newbies to react. We can change it to the component name maybe. @aquibbaig

dashboard/v1.0/.prettierrc.json
```diff
@@ -3,7 +3,8 @@
         {
             "files": [".prettierrc", ".babelrc", ".eslintrc", ".stylelintrc"],
             "options": {
-                "parser": "json"
+                "parser": "json",
+                "tabWidth": 2
```
Harkishen-Singh: Please make sure that this has effect in the CI tests. They should run strictly against these rules and fail on a single out. Else, this would be pointless.

dashboard/v1.0/src/App.tsx
```diff
@@ -1,20 +1,13 @@
...
-        <Sidebar />
-      </div>
-      <div className="inner-component">
-        <Header />
-        <Navigator />
-      </div>
-    </div>
+    <GlobalLayout>
+      <Navigator />
+    </GlobalLayout>
```
Harkishen-Singh: Could you please tell why is this necessary?
Also, `GlobalLayout` should be changed to something more relatable like `Wrap` or something related to the topic. _(I am not good at naming)_

dashboard/v1.0/src/components/layouts/GlobalLayout.tsx
Harkishen-Singh: Agreed. How about `root`? as a root component in the react-dom.

aquibbaig: Thanks @ganeshpatro321!

aquibbaig:
@Harkishen-Singh,
> App.js needs to be singular and free of all nested components, only subscribed to global stores and persist configurations. Common components for each page should come under a global layout component. 

Having a layout simplifies much of our codebase, freeing App.js only to things that it needs. We may need to add other components such as ContainerQuery, Persistors, etc. It's a common practice for `React` developers.

dashboard/v1.0/src/App.tsx
aquibbaig: It's a layout that uses composition. So `GlobalLayout` is perfect. We can use `BasicLayout` otherwise.

Harkishen-Singh: I just want to avoid `Global` from naming. Why not give the name related to something you want to implement?

aquibbaig: `BasicLayout` or `DefaultLayout` will be good?

aquibbaig: `BasicLayout` or `DefaultLayout` will be good?

aquibbaig: `BasicLayout` or `DefaultLayout` will be good?

aquibbaig: cool

dashboard/v1.0/src/components/layouts/BaseLayout.tsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import Header from './Header';
+import Sidebar from './Sidebar';
+
+export default function BaseLayout(props) {
+  return (
+    <>
```
Harkishen-Singh: We can remove these. Its being handled by `div`

dashboard/v1.0/src/components/layouts/BaseLayout.tsx
```diff
@@ -0,0 +1,19 @@
...
+      <div className="App">
+        <div className="sidebar">
+          <Sidebar />
+        </div>
+        <div className="inner-component">
+          <Header />
+          {props.children}
+        </div>
+      </div>
+    </>
```
Harkishen-Singh: same.

PR #173, opened 2020-03-07
Initial setup for material-ui in bench-routes

aquibbaig:
@ganeshpatro321 @Harkishen-Singh Drop a review.


dashboard/v1.1/src/App.tsx
```diff
@@ -1,8 +1,32 @@
...
+    fontSize: 12
+  }
+});
+theme = responsiveFontSizes(theme);
 
 function App() {
-  return <div className="App">Bench-routes ui</div>;
+  return (
+    <ThemeProvider theme={theme}>
+      <div className="App">Bench-routes ui</div>
```
Harkishen-Singh: can we split this into multiple lies? would appeal good.

aquibbaig: Can't because `prettier-fix` automatically reverts it back.

PR #238, opened 2020-05-04
Extended api support for modules

muskankhedia:
Signed-off-by: muskankhedia <muskan.khedia2000@gmail.com>

Presently the API support was only for `sys` type of files. In this PR the support has been extended for `ping`, `jitter` and `floodping` types of files.

### DIFF
--- a/src/lib/modules/ping/ping.go
+++ b/src/lib/modules/ping/ping.go
@@ -27,7 +27,7 @@ type TestInterval struct {
 	Duration int64
 }
 
-// Response is used to decode the tsdb while sending data in the API
+// Response is used to decode the tsdb blocks to data points that supports JSON encoding.
 type Response struct {
 	Min  string `json:"minValue"`
 	Avg  string `json:"avgValue"`
--- a/src/lib/modules/ping/ping.go
+++ b/src/lib/modules/ping/ping.go
@@ -27,6 +27,14 @@ type TestInterval struct {
 	Duration int64
 }
 
+// Response is used to decode the tsdb while sending data in the API
+type Response struct {
+	Min  string `json: "minValue"`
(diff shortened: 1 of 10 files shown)


src/lib/modules/ping/ping.go
```diff
@@ -27,6 +27,14 @@ type TestInterval struct {
 	Duration int64
 }
 
+// Response is used to decode the tsdb while sending data in the API
```
Harkishen-Singh, line 30: `response is used to decode the tsdb blocks to data points that supports JSON encoding.`

Please change others according to this as well.

src/lib/utils/decode/module-decoding.go
```diff
@@ -0,0 +1,68 @@
...
+	"fmt"
+	"strings"
+
+	"github.com/zairza-cetb/bench-routes/src/lib/modules/jitter"
+	"github.com/zairza-cetb/bench-routes/src/lib/modules/ping"
+	"github.com/zairza-cetb/bench-routes/src/metrics/system"
+)
+
+// System converts the block into Response type for easily http based
+// response.
```
Harkishen-Singh, line 13: This needs to be updated.

src/lib/utils/decode/module-decoding.go
```diff
@@ -0,0 +1,68 @@
+package decode
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/zairza-cetb/bench-routes/src/lib/modules/jitter"
+	"github.com/zairza-cetb/bench-routes/src/lib/modules/ping"
+	"github.com/zairza-cetb/bench-routes/src/metrics/system"
```
Harkishen-Singh, line 9: implement this package as a form of an interface as it was earlier in the decoder.

Issue #226, opened 2020-04-11
Go profiling in bench-routes


Harkishen-Singh:
We need to support profiling of go. This is necessary to get an insight into the mechanism happening in the go-routines and in the heap.

akshaysingh5:
Can you please elaborate on the issue and give some initial guidance, I would like to work on this issue.,

Harkishen-Singh:
Sure @akshaysingh5 . The best place would be the original docs which are rather simple to understand, https://golang.org/pkg/net/http/pprof/ and along with articles on the web that you can search for in-depth knowledge.

aquibbaig:
@akshaysingh5 Are you working on this issue?

Harkishen-Singh:
@akshaysingh5 any updates?

aquibbaig:
I guess there are no updates on this. @akshaysingh5 Feel free to take up any other issues.

aquibbaig:
Solved via #252

PR #252, opened 2020-05-19
Add support for pprof in bench-routes

aquibbaig:
Aimed at fixing #226.
Exposes specific http endpoints to profiling using "pprof". There are a load bunch of other
things we can add support to. Need a discussion on that. By default I have added the following profiling options:
- go routine
- thread create
- block
- cmdline

### DIFF
--- a/src/main.go
+++ b/src/main.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"net/http/pprof"
 	"os"
 	"os/signal"
 	"reflect"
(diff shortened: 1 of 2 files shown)


src/main.go
```diff
@@ -388,6 +389,28 @@ func main() {
...
+		The package initialization registers it as /debug/pprof/cmdline. **/
+		router.HandleFunc("/debug/pprof/cmdline", pprof.Cmdline)
+		/** Profile responds with the pprof-formatted cpu profile.
+		Profiling lasts for duration specified in seconds GET parameter,
+		or for 30 seconds if not specified. The package initialization registers it as /debug/pprof/profile. **/
+		router.HandleFunc("/debug/pprof/profile", pprof.Profile)
+		/** Symbol looks up the program counters listed in the request, responding
+		with a table mapping program counters to function names.
+		The package initialization registers it as /debug/pprof/symbol. **/
+		router.HandleFunc("/debug/pprof/symbol", pprof.Symbol)
```
Harkishen-Singh, line 413: So many details. Better to have these in another file/directory that does the sole purpose of monitoring bench-routes and its behaviour when under load.This area can be expanded once bench-routes is in production cases.

PR #276, opened 2020-06-26
Support for post request in monitoring screen

Harkishen-Singh:
Signed-off-by: Harkishen-Singh <harkishensingh@hotmail.com>

### DIFF
--- a/local-config.yml
+++ b/local-config.yml
@@ -107,3 +107,14 @@ routes:
   headers: []
   params: []
   body: []
+- method: POST
+  url: https://reqres.in/api/register
+  headers: []
+  params: []
+  body:
+  - name: ""
+    value: ""
+  - name: email
+    value: eve.holt@reqres.in
+  - name: password
+    value: pistol
--- a/src/lib/modules/monitor/monitor.go
+++ b/src/lib/modules/monitor/monitor.go
@@ -1,8 +1,6 @@
 package monitor
 
 import (
-	// "io/ioutil"
-	// "math"
 	"strconv"
 	"sync"
 	"time"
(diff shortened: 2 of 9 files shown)


local-config.yml
```diff
@@ -107,3 +107,14 @@ routes:
...
+  url: https://reqres.in/api/register
+  headers: []
+  params: []
+  body:
+  - name: ""
+    value: ""
+  - name: email
+    value: eve.holt@reqres.in
+  - name: password
+    value: pistol
```
aquibbaig, line 120: why not an array?

src/lib/modules/monitor/monitor.go
```diff
@@ -136,10 +125,19 @@ func (ps *Monitor) responseDelay(wg *sync.WaitGroup, route parser.Route) {
...
 
-	c := make(chan utils.Response)
-	responseObject := routeDispatcher(route, c)
+	response := make(chan string)
+	req := request.New(route.URL, request.ToMap(route.Header), request.ToMap(route.Params), request.ToMap(route.Body))
+	// responseObject := routeDispatcher(route, c)
+	stamp := time.Now()
+	go req.Send(request.MethodUintPresentation(route.Method), response)
+	resp := <-response
+	c := utils.Response{
```
aquibbaig, line 134: rename this variable to something more appropriate because c is reserved for "channels"

local-config.yml
aquibbaig: If we do not opt to store this as an array, we would be converting the response each time to an iterable "name" -> "value"

src/lib/modules/monitor/monitor.go
```diff
@@ -1,8 +1,8 @@
 package monitor
 
 import (
-	"io/ioutil"
-	"math"
+	// "io/ioutil"
+	// "math"
```
aquibbaig, line 5: Remove comments if not required

local-config.yml
Harkishen-Singh: It is an array. Non-empty arrays in yaml are represented in form of:

```
a:
- para_1:
  para_2:
  ..
- para_3:
  para_4
```

that corresponds to `[{para_1, para_2}, {para_3, para_4}]`

src/lib/modules/monitor/monitor.go
Harkishen-Singh: apparently, `c` is not required as well.i.e., without the use of a var, we can go ahead as well.

src/lib/modules/monitor/monitor.go
Harkishen-Singh: done

Harkishen-Singh:
@aquibbaig implemented!

PR #278, opened 2020-07-18
Support for config screen and various go improvements

aquibbaig:
Works to be done in this PR:
- [x] Replace the URL in the URI for each route
- [x] Delete routes from config screen
- [x] Alert when service is unreachable


src/lib/api/api.go
```diff
@@ -467,6 +468,52 @@ func (a *API) ModifyIntervalDuration(w http.ResponseWriter, r *http.Request) {
...
+		Action      string `json:"action"`
+		NewRoute    string `json:"newRoute"`
+		ActualRoute string `json:"actualRoute"`
+	}
+	decoder := json.NewDecoder(r.Body)
+	if err := decoder.Decode(&req); err != nil {
+		panic(err)
+	}
+	switch req.Action {
+	case "edit-route":
```
Harkishen-Singh, line 486: since `route` is expected to be redundant here, how about simply `edit`, `delete`, etc.?

src/lib/api/api.go
```diff
@@ -467,6 +468,50 @@ func (a *API) ModifyIntervalDuration(w http.ResponseWriter, r *http.Request) {
...
+			if route.URL == req.ActualRoute {
+				a.config.Config.Routes = append(a.config.Config.Routes[:i], a.config.Config.Routes[i+1:]...)
+			}
+		}
+	}
+	_, err := a.config.Write()
+	if err != nil {
+		a.ResponseStatus = http.StatusText(400)
+		panic(err)
+	}
```
Harkishen-Singh, line 509:
```

	if _, err := a.config.Write(); err != nil {
		a.ResponseStatus = http.StatusText(400)
		panic(err)
	}
```
also, if you panic, there won't be a `ResponseStatus` sent to the UI. So, rather than panicking, go with logging an error so that it sends the UI.

src/lib/api/api.go
```diff
@@ -97,6 +97,7 @@ func (a *API) Register(router *mux.Router) {
 	router.HandleFunc("/get-config-intervals", a.GetConfigIntervals)
 	router.HandleFunc("/get-config-routes", a.GetConfigRoutes)
 	router.HandleFunc("/get-route-time-series", a.TSDBPathDetails)
+	router.HandleFunc("/modify-config-routes", a.ModifyConfigRoutes)
```
Harkishen-Singh, line 100: how about `/update-routes`?

aquibbaig:
@Harkishen-Singh Docker service is failing for some reason

Harkishen-Singh:
After editing the route from the config screen, the changes are not visible in the monitoring screen. Consider reloading the application by passing `*a.reloadConfigURLs <- struct{}`. That should solve this issue.

As far as tests are concerned, the secondary libraries seem to break. We might have to switch the lib that is breaking the code if it does not pass on the second run.

aquibbaig:
@Harkishen-Singh do you mean using a modal where the route and it's options are shown ? And user can edit each one of them?

Harkishen-Singh:
> @Harkishen-Singh do you mean using a modal where the route and it's options are shown ? And user can edit each one of them?

Yes, at least according to the discussion we had.

aquibbaig:
@Harkishen-Singh that way we need to disable row customisation of mui and need to design a custom editing design

Harkishen-Singh:
> @Harkishen-Singh that way we need to disable row customisation of mui and need to design a custom editing design

If that's the only way, then we have to; if not in this PR, then let's have an issue with TODO as this. 

Btw, does the current implementation support editing of params/headers/body?

aquibbaig:
> > @Harkishen-Singh that way we need to disable row customisation of mui and need to design a custom editing design
> 
> If that's the only way, then we have to; if not in this PR, then let's have an issue with TODO as this.
> 
> Btw, does the current implementation support editing of params/headers/body?

It doesn't. It can only allow to edit what's on the table, which in our case is the route name and the methods

Harkishen-Singh:
Okay, then anyway we need yo make a custom one in order to match the UI and also this. That should be the correct way.

What do you say?

aquibbaig:
> Okay, then anyway we need yo make a custom one in order to match the UI and also this. That should be the correct way.
> 
> What do you say?

I guess deleting works fine as it is. It is editing that we must custom build

aquibbaig:
@Harkishen-Singh TODO are all of the interface exports and type cleanups. Please try to review them once, I will change them tomortow anyways

dashboard/v1.1/src/pages/Config/Config.tsx
```diff
@@ -42,84 +55,77 @@ const Config: FC<{}> = () => {
...
+            method: route.Method,
+            body: route.Body,
+            headers: route.Header,
+            params: route.Params
+          }
+        ]);
+      }
+    });
+    setConfigRoutes(configRoutes);
+  };
```
Harkishen-Singh, line 128: Let's have these long functions to a different file in the same directory as `handles.ts`.

dashboard/v1.1/src/pages/Config/Config.tsx
```diff
@@ -42,84 +55,77 @@ const Config: FC<{}> = () => {
...
 
   const handleToggle = (intervalName: string) => {
```
Harkishen-Singh, line 130: This too in the `handles.ts`

dashboard/v1.1/src/pages/Config/Config.tsx
```diff
@@ -143,77 +149,145 @@ const Config: FC<{}> = () => {
...
+                tooltip: 'Edit Route',
+                onClick: (event, rowData: TableRowData) => {
+                  setSelectedRow({
+                    route: rowData.route,
+                    options: configRoutes.get(rowData.route)
+                  });
+                  setEditModalOpen(!editModalOpen);
+                }
+              }
+            ]}
```
Harkishen-Singh, line 242: please see whether this can be declared in `handles.ts` and then imported here.

dashboard/v1.1/src/pages/Config/Config.tsx
```diff
@@ -143,77 +149,145 @@ const Config: FC<{}> = () => {
...
+                        }
+                      });
+                      setConfigRoutes(configRoutes);
+                    },
+                    e => {
+                      console.error(e);
+                    }
+                  );
+              }
+            }}
```
Harkishen-Singh, line 287: Let's have this entire function in `handles.ts`. This will simplify `Config.tsx` by a huge margin.
Also, use the interface `route` at `data.forEach(route => {` and then instead of `[<...>]`, lets use `.` operator.

dashboard/v1.1/src/pages/Config/Services.ts
```diff
@@ -0,0 +1,67 @@
...
+          params: route.Params
+        }]);
+    } else {
+      configRoutes.set(uri, [
+        ...configRoutes.get(uri), {
+          method: route.Method,
+          body: route.Body,
+          headers: route.Header,
+          params: route.Params
+        }]);
```
Harkishen-Singh, line 63: Like the ones used here.

dashboard/v1.1/src/pages/Config/components/EditModal.tsx
```diff
@@ -0,0 +1,127 @@
...
+                    variant="outlined"
+                    size="small"
+                    label={selectedRoute.route}
+                    clickable
+                    color="primary"
+                  />
+                }
+              </div>
+            ) : (
+              <div> </div>
```
Harkishen-Singh, line 71: Please confirm if this is needed. Else `null` would be much preferred.

dashboard/v1.1/src/pages/Input/Input.tsx
```diff
@@ -14,6 +14,17 @@ import DialogContent from '@material-ui/core/DialogContent';
...
+import PropTypes from 'prop-types';
+
+interface InputScreenProps {
+  screenType: string | undefined;
+  params: { Name: string; Value: string }[];
+  headers: { OfType: string; Value: string }[];
+  route: string;
+  body: { Name: string; Value: string }[];
+  method: string;
+  updateCurrModal: (routes: any) => void;
```
Harkishen-Singh, line 26: `updateCurrModal` => `updateCurrentModal`

dashboard/v1.1/src/pages/Input/Input.tsx
```diff
@@ -53,6 +72,37 @@ const Input: FC<{}> = () => {
   const [open, setOpen] = useState(false);
   const [showResponseButton, setShowResponseButton] = useState<boolean>(false);
 
+  useEffect(() => {
+    if (screenType === 'configScreen') {
```
Harkishen-Singh, line 76: `configScreen` => `config-screen`

dashboard/v1.1/src/pages/Input/Input.tsx
```diff
@@ -53,6 +72,37 @@ const Input: FC<{}> = () => {
...
+          key: param.Name,
+          value: param.Value
+        });
+      });
+      body?.forEach(val => {
+        bodyValues.push({
+          key: val.Name,
+          value: val.Value
+        });
+      });
```
Harkishen-Singh, line 91: Both of these look similar. A single function with 2 times call will be better to use.

dashboard/v1.1/src/pages/Input/Input.tsx
```diff
@@ -82,6 +132,80 @@ const Input: FC<{}> = () => {
     setApplyParams(false);
     setApplyBody(false);
   };
+  const testAndEdit = () => {
+    const params = {};
```
Harkishen-Singh, line 136: Can the length of this function be reduced? If not, please consider having a different file for `testAndEdit` and `testURL` to have this file short.

src/lib/api/api.go
```diff
@@ -467,6 +467,85 @@ func (a *API) ModifyIntervalDuration(w http.ResponseWriter, r *http.Request) {
...
+			a.mux.Lock()
+			a.config.Config.Routes = append(a.config.Config.Routes[:i], a.config.Config.Routes[i+1:]...)
+			a.mux.Unlock()
+			break
+		}
+	}
+	_, err := a.config.Write()
+	if err != nil {
+		a.ResponseStatus = http.StatusText(400)
+	}
```
Harkishen-Singh, line 543:
```
if _, err := a.config.Write(); err != nil {
	a.ResponseStatus = http.StatusText(400)
}
```

dashboard/v1.1/src/pages/Config/Config.tsx
aquibbaig: I'd rather not do that because these are state manipulations, let's keep them in here for now

dashboard/v1.1/src/pages/Config/Config.tsx
aquibbaig: As an additional prop here, I'd suggest this to be here as it is

dashboard/v1.1/src/pages/Config/Config.tsx
Harkishen-Singh: The file is being too long for simple functions. Thats my point.

PR #281, opened 2020-07-24
Support for config screen and various go improvements

aquibbaig:
For the new config screen.
Works to be done in this PR:
- [x] Replace the URL in the URI for each route
- [x] Delete routes from config screen
- [x] Alert when service is unreachable


aquibbaig:
@Harkishen-Singh Will make a new PR with changes. I figured out what was wrong

Harkishen-Singh:
@aquibbaig sure, that should be good.

PR #282, opened 2020-07-24
Adds support to edit/delete routes from config screen along with various go enhancements

aquibbaig:
Using all new config screen, it is pretty easy to edit
and delete routes stored in local-config and add new routes
to monitoring in bench routes


aquibbaig:
> Looks good to me. Expecting that a basic run on this is already done by you!

I gave it a run once

PR #297, opened 2020-10-16
create a systemd file

aquibbaig:
Creates a systemd-service file.
- this file will be copied to the `/etc/systemd` folder when we install bench routes.
- ensure that all dependencies are properly installed before starting the service.
- start bench routes and dashboard in one service.


aquibbaig:
NOTE: Docker would fail because of "systemctl" not being defined in the container

aquibbaig:
Done in the alpha release.

PR #298, opened 2020-11-08
move tsdb as v1 milestone

aquibbaig:
We'd  be working as our new tsdb model in this branch. I updated a lot of dependencies too as a part of this PR, which hopefully doesn't break a lot of things :(


aquibbaig:
Closing due to the fact that we updated dependencies in another PR, and changed our continuous integration

PR #303, opened 2020-12-04
implemented hash-table

muskankhedia:
Signed-off-by: muskankhedia <muskan.khedia2000@gmail.com>

Implemented the hash table to save the hash value for every type and URL in the config file.


aquibbaig:
@muskankhedia Since this is a collaborative effort, why not work on a single branch, rather than having to manage over multiple branches. The other way around is for us to have a separate repository which can act as a module until its ready and can be merged in the main repository.

src/main.go
```diff
@@ -3,6 +3,7 @@ package main
 import (
 	"encoding/json"
 	"fmt"
+	"github.com/zairza-cetb/bench-routes/tsdb"
```
Harkishen-Singh, line 6: THis and the line below, both can be kept aside.

```suggestion
import (
	"encoding/json"
	"fmt"
	"net/http"
	"os"
	
	"github.com/zairza-cetb/bench-routes/tsdb"
	"github.com/zairza-cetb/bench-routes/tsdb/v1"
```

src/main.go
```diff
@@ -38,17 +39,19 @@ var (
 	defaultScrapeTime           = time.Second * 3
 	systemMetricsPath           = "storage/system.json"
 	journalMetricsPath          = "storage/journal.json"
+	filesPath                   = "storage/000000"
```
Harkishen-Singh, line 42: This shouldn' be hardcoded. also, the `000000` should be kept as int and can be convertable to string so that we can do arithmetic operations on it.

src/main.go
```diff
@@ -139,6 +166,10 @@ func main() {
 					chainSet.Register(fmt.Sprintf("%s-monitor", uHash), matrix[hash].MonitorChain)
 				}
 			}
+			e := tsdb.SaveHashTableToHDD(filesPath, &hashListData)
+			if e != nil {
+				panic(e)
+			}
```
Harkishen-Singh, line 172:
```suggestion
			if err := tsdb.SaveHashTableToHDD(filesPath, &hashListData); err != nil {
				panic(err)
			}
```

src/main.go
```diff
@@ -313,7 +344,7 @@ func setDefaultServicesState(configuration *parser.Config) {
 }
 
 // URLHash hashes the passed route to a unique value.
-func URLHash(route parser.Route) string {
+func URLHash(route parser.Route, checkType string) string {
```
Harkishen-Singh, line 347: Being in the main, this function should not be exported.

tsdb/db.go
```diff
@@ -0,0 +1,86 @@
...
+func saveToHDD(path string, data []byte) error {
+	checkAndCreatePath(path)
+	e := ioutil.WriteFile(path, data, 0755)
+	if e != nil {
+		return e
+	}
+	return nil
+}
+
+//SaveHashTableToHDD saves the hashtable to HDD
```
Harkishen-Singh, line 68:
```suggestion
// SaveHashTable saves the hashtable to the disk.
```

Harkishen-Singh: I also think that calling this as **hash.*** is confusing. This should be something like an index or indextable.

tsdb/db.go
```diff
@@ -0,0 +1,86 @@
...
+func SaveHashTableToHDD(path string, data *map[string]*HashList) error {
+	checkAndCreatePath(path)
+	res := formatToStoreData(data)
+	write := []byte(res)
+	e := ioutil.WriteFile(path, write, 0755)
+	if e != nil {
```
Harkishen-Singh, line 74: one liner please

tsdb/db.go
```diff
@@ -0,0 +1,86 @@
...
+	str := string(res)
+	return &str, nil
+}
+
+func checkAndCreatePath(path string) {
+	array := strings.Split(path, "/")
+	array = array[:len(array)-1]
+	path = strings.Join(array, "/")
+	_, err := os.Stat(path)
+	if err != nil {
```
Harkishen-Singh, line 51: one liner please

Harkishen-Singh (changes requested):

Harkishen-Singh:
@aquibbaig , the `tsdb-v2` is the master branch of tsdb v2. Please make your respective branches to push changes to this brnahc only via pull request. We should not be pushing directly to the `tsdb-v2`.

src/main.go
aquibbaig: I'd like to know the motivation behind this pathname?

src/main.go
```diff
@@ -106,7 +109,7 @@ func main() {
 			conf.Refresh()
 			p := time.Now()
 			for _, r := range conf.Config.Routes {
-				hash := URLHash(r)
+				hash := URLHash(r, "")
```
aquibbaig, line 112: Why do we pass an empty string here, if its deliberate we need to mention that in the comments

aquibbaig (changes requested):

src/main.go
muskankhedia: There is no particular reason for that. I have just named in for development purpose, we can change it later.

tsdb/db.go
```diff
@@ -0,0 +1,86 @@
...
+	res := formatToStoreData(data)
+	write := []byte(res)
+	e := ioutil.WriteFile(path, write, 0755)
+	if e != nil {
+		return e
+	}
+	return nil
+}
+
+func formatToStoreData(data *map[string]*HashList) string {
```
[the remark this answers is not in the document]
muskankhedia: Any suggestions?

src/main.go
```diff
@@ -115,6 +118,30 @@ func main() {
 						pathMonitor   = fmt.Sprintf("%s/chunk_monitor_%s.json", pathMonitoring, hash)
 					)
 					uHash := utils.GetHash(filters.HTTPPingFilterValue(r.URL))
+					hashping := URLHash(r, "ping")
```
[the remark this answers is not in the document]
muskankhedia: It's not possible to make changes to the `matrix` or `targetMachineCalc` as they are quite complex. Also after the entire implementation of hashlist and new tsdb we can remove the previous implementation entirely.

src/main.go
```diff
@@ -115,6 +118,30 @@ func main() {
...
+					hashjitter := URLHash(r, "jitter")
+					hashfp := URLHash(r, "flood-ping")
+					hashreqres := URLHash(r, "reqres")
```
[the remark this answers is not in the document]
muskankhedia: Sure. I have made changes accordingly.

src/main.go
muskankhedia: Sure.

src/main.go
Harkishen-Singh: I think this still needs to be updated? @muskankhedia

Harkishen-Singh (approved):
lgtm to me

Harkishen-Singh (approved):

src/main.go
```diff
@@ -29,26 +28,31 @@ import (
...
 
 var (
 	port                        = ":9990" // default listen and serve at 9090
 	enableProcessCollection     = false   // default collection of process metrics in host of bench-routes
 	processCollectionScrapeTime = time.Second * 5
 	defaultScrapeTime           = time.Second * 3
+	storageFile                 = "000000"
 	systemMetricsPath           = "storage/system.json"
 	journalMetricsPath          = "storage/journal.json"
+	filesPath                   = "storage/" + storageFile
```
chetanyakan, line 43: nit - You can use the variable `storageDir` defined on line 63

src/main.go
```diff
@@ -115,6 +119,16 @@ func main() {
 						pathMonitor   = fmt.Sprintf("%s/chunk_monitor_%s.json", pathMonitoring, hash)
 					)
 					uHash := utils.GetHash(filters.HTTPPingFilterValue(r.URL))
+					types := [4]string{"ping", "jitter", "flood-ping", "req-res"}
+					for i := 0; i < 4; i++ {
+						fmt.Println(i, types[i])
```
chetanyakan, line 124: Remove if not needed.
We should use `log.Infoln()` if needed.

src/main.go
Harkishen-Singh: @chetanyakan this will be rewritten by binary tsdb PR, so I think that's why we arent updating this PR. It awesome to receive PR reviews. Thank you for that.

PR #304, opened 2020-12-15
Migrate to Github actions

aquibbaig:
This PR would fix our golang tests for the mac environment


aquibbaig:
@Harkishen-Singh We need to seggregate our tests based on whether we are testing for golang or dashboard/javascript. After that we can test for docker build and also consequent releases

muskankhedia (approved):
LGTM!

PR #328, opened 2021-02-16
simplification of main goroutine

aquibbaig:
This PR simplifies the main goroutine.

### DIFF
--- a/src/lib/api/api.go
+++ b/src/lib/api/api.go
@@ -11,6 +11,10 @@ import (
 	"sync"
 	"time"
 
+	"github.com/gorilla/mux"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"github.com/prometheus/common/log"
+
 	config "github.com/bench-routes/bench-routes/src/lib/config"
 	"github.com/bench-routes/bench-routes/src/lib/modules/jitter"
 	"github.com/bench-routes/bench-routes/src/lib/modules/monitor"
@@ -20,9 +24,6 @@ import (
 	"github.com/bench-routes/bench-routes/src/lib/utils/brt"
 	"github.com/bench-routes/bench-routes/tsdb"
 	"github.com/bench-routes/bench-routes/tsdb/querier"
-	"github.com/gorilla/mux"
-	"github.com/prometheus/client_golang/prometheus/promhttp"
-	"github.com/prometheus/common/log"
 )
 
 const (
@@ -32,14 +33,13 @@ const (
 
 // API type for implementing the API interface.
 type API struct {
-	ResponseStatus      string
-	Data, Services      interface{}
-	Matrix              *map[string]*utils.BRMatrix
-	configurationPath   string
-	config              *config.Config
-	reloadConfigURLs    chan struct{}
-	receiveFinishSignal chan struct{}
-	mux                 sync.RWMutex
+	ResponseStatus    string
+	Data, Services    interface{}
+	Matrix            *map[string]*utils.BRMatrix
+	configurationPath string
+	config            *config.Config
+	reloadConfigURLs  chan struct{}
+	mux               sync.RWMutex
 }
 
 type inputRequest struct {
@@ -57,16 +57,14 @@ func New(
 	config *config.Config,
 	configPath string,
 	services interface{},
-	reload,
-	done chan struct{},
+	reload chan struct{},
 ) *API {
 	return &API{
-		Matrix:              matrix,
-		Services:            services,
-		configurationPath:   configPath,
-		config:              config,
-		reloadConfigURLs:    reload,
-		receiveFinishSignal: done,
+		Matrix:            matrix,
+		Services:          services,
+		configurationPath: configPath,
+		config:            config,
+		reloadConfigURLs:  reload,
 	}
 }
 
@@ -341,7 +339,6 @@ func (a *API) AddRouteToMonitoring(w http.ResponseWriter, r *http.Request) {
 	a.reloadConfigURLs <- struct{}{}
 	a.Data = "success"
 	a.send(w, a.marshalled())
-	<-a.receiveFinishSignal
 }
 
 // TSDBPathDetails responds with the path details that will be used for
@@ -512,7 +509,6 @@ func (a *API) UpdateRoute(w http.ResponseWriter, r *http.Request) {
 		a.ResponseStatus = http.StatusText(400)
 	}
 	a.reloadConfigURLs <- struct{}{}
-	<-a.receiveFinishSignal
 	a.ResponseStatus = http.StatusText(200)
 	a.Data = a.config.Config.Routes
 	a.send(w, a.marshalled())
--- a/src/main.go
+++ b/src/main.go
@@ -13,6 +13,11 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/gorilla/mux"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/common/log"
+	"github.com/rs/cors"
+
 	"github.com/bench-routes/bench-routes/src/lib/api"
 	parser "github.com/bench-routes/bench-routes/src/lib/config"
 	"github.com/bench-routes/bench-routes/src/lib/filters"
@@ -25,10 +30,6 @@ import (
 	"github.com/bench-routes/bench-routes/src/metrics/process"
 	sysMetrics "github.com/bench-routes/bench-routes/src/metrics/system"
 	"github.com/bench-routes/bench-routes/tsdb"
-	"github.com/gorilla/mux"
-	"github.com/prometheus/client_golang/prometheus"
-	"github.com/prometheus/common/log"
-	"github.com/rs/cors"
 )
 
 var (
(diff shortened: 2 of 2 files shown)


.github/workflows/build-linux.yml
```diff
@@ -24,7 +24,6 @@ jobs:
           $(go env GOPATH)/bin/golangci-lint run
           echo "perfect linting standards! moving ahead ..."
           echo "checking go format ..."
-          diff -u <(echo -n) <(gofmt -d ./src ./tsdb)
```
aquibbaig, line 27: Tasks were failing due to the output emitted from diff.

.github/workflows/build-mac.yml
```diff
@@ -7,7 +7,7 @@ on: [pull_request, push]
 jobs:
   build:
     name: build
-    runs-on: macOS-11.0
+    runs-on: macos-latest
```
aquibbaig, line 10: They have removed the label from support. `macos-latest` should work just fine

src/main.go
```diff
@@ -293,10 +308,8 @@ func main() {
 		log.Infof("Alive %d goroutines after cleaning up.\n", runtime.NumGoroutine())
 		os.Exit(0)
 	}()
-	log.Infoln(http.ListenAndServe(port, cors.Default().Handler(router)).Error())
-	// keep the below line to the end of file so that we ensure that we give a confirmation message only when all the
-	// required resources for the application is up and healthy.
 	log.Infoln("Bench-routes is up and running")
+	log.Infoln(http.ListenAndServe(port, cors.Default().Handler(router)).Error())
```
aquibbaig, line 312: Code is always blocking on this line...

src/main.go
```diff
@@ -3,6 +3,9 @@ package main
 import (
 	"encoding/json"
 	"fmt"
+	"github.com/rs/cors"
+	"github.com/zairza-cetb/bench-routes/src/lib/filters"
+	"github.com/zairza-cetb/bench-routes/tsdb"
```
Harkishen-Singh, line 8: please fix the format of importing. (same between default imports and third party imports)

src/main.go
```diff
@@ -99,14 +102,22 @@ func main() {
 		endpointMetrics.StatusCode,
 		endpointMetrics.MonitorCount,
 	)
+	// go routine that reloads the entire map(cMap).
 	go func() {
-		for {
-			<-reload
-			log.Infoln("reloading...")
+		<-start
```
Harkishen-Singh, line 107: sorry, i didnt get why is <-start here?

src/main.go
```diff
@@ -138,13 +151,15 @@ func main() {
 					}
 					chainSet.Register(fmt.Sprintf("%s-monitor", uHash), matrix[hash].MonitorChain)
 				}
+				// If the route is already present in the matrix,
+				// the route is already monitored, we just continue
+				// normal execution.
 			}
 			log.Infoln("initialization time: " + time.Since(p).String())
-			done <- struct{}{}
```
Harkishen-Singh, line 143: If this is removed, how will the reloading caller know that the reload process is completed?

src/main.go
Harkishen-Singh: It is meant to be blocked here. Do we not want to block it?

Harkishen-Singh (changes requested):

src/main.go
aquibbaig: You could do a `log.Infoln`

src/main.go
aquibbaig: The `start` channel simply triggers the goroutine on startup, indicating that the goroutine has started

src/main.go
aquibbaig: Of course we would. But printing "bench routes is up and running" after this line would never be called so I moved it up

.github/workflows/build-linux.yml
aquibbaig: What about explicitely using some git hooks to format it before each commit.

src/main.go
Harkishen-Singh: Sorry, but the `<-` is for receiving a signal, not sending a signal. It can block there.

src/main.go
Harkishen-Singh: I mean to say, the caller means an API. The API needs to send back success or failure. The success is only when the reload success signal is sent. Logging wouldn't help that.

go.mod
```diff
@@ -4,20 +4,20 @@ go 1.12
```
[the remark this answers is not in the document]
aquibbaig: Dropped the changes to `go.mod` and `go.sum`

src/main.go
aquibbaig: Interesting that it is not caught by the editor

src/main.go
aquibbaig: Done

src/lib/api/api.go
```diff
@@ -339,9 +340,10 @@ func (a *API) AddRouteToMonitoring(w http.ResponseWriter, r *http.Request) {
 		),
 	)
 	a.reloadConfigURLs <- struct{}{}
+	// wait for reload to finish
+	<-a.receiveFinishSignal
```
Harkishen-Singh, line 344: The reason this was kept in 344 is, routes can take a lot of time to reload. We do not have to wait for that to happen. If we wait, the only reason would have been an error possibility which I don't see happening (since the reloading goroutine does not return any error as there isn't anything to cause an error).

src/main.go
```diff
@@ -70,11 +72,12 @@ func main() {
 	} else if len(os.Args) > 1 {
 		port = ":" + os.Args[1]
 	}
-	log.Infoln("initializing...")
+	log.Infoln("initializing bench routes...")
```
Harkishen-Singh, line 75:
```suggestion
	log.Infoln("initializing bench-routes...")
```

src/main.go
```diff
@@ -70,11 +72,12 @@ func main() {
...
 	conf = parser.New(configurationFilePath)
 	conf.Load().Validate()
 	setDefaultServicesState(conf)
 	intervals := conf.Config.Interval
+	// Loads workers on to the memory
```
Harkishen-Singh, line 80:
```suggestion
	// Loads workers on to the memory.
```
I would rather say "Initialise workers" but please feel free to ignore this.

src/main.go
```diff
@@ -42,15 +43,16 @@ var (
...
+	// Filters out identical routes that arrive to be monitored.
 	matrix   = make(map[string]*utils.BRMatrix)
 	reload   = make(chan struct{})
 	done     = make(chan struct{})
 	conf     *parser.Config
 	chainSet = tsdb.NewChainSet(tsdb.FlushAsTime, time.Second*300)
 	// targetMachineCalc contains calculations that are machine/vm/load-balancer
 	// specific. These involve use of IP addresses/Domain names respectively.
+	// We use this to keep track of host names because all the routes that share
+	// the same hostname will have the same ping, jitter and flood-ping.
```
Harkishen-Singh, line 55: This is a good replacement of comment, but I do feel the 45 & 46 line comment should be added here. It is useful for contributors to know why this behaviour was done. Helps in understanding the already complex map.

src/main.go
```diff
@@ -99,14 +102,21 @@ func main() {
 		endpointMetrics.StatusCode,
 		endpointMetrics.MonitorCount,
 	)
+	// go routine that reloads the entire map(cMap).
 	go func() {
-		for {
-			<-reload
-			log.Infoln("reloading...")
+		log.Infoln("Initiated monitoring your routes...")
```
Harkishen-Singh, line 107: Do we need this? (the below reloading seems enough. maybe change `reloading` => `refreshing new routes in memory`)

src/main.go
```diff
@@ -123,10 +133,12 @@ func main() {
 							FPing:    tsdb.NewChain(pathFloodPing).Init(),
 							Metrics:  targetMachineMetrics,
 						}
+						// commits new chain after every interval.
```
Harkishen-Singh, line 136:
```suggestion
						// Register new chains so that they get committed if they contain any new blocks, in regular intervals.
```

src/main.go
```diff
@@ -123,10 +133,12 @@ func main() {
...
 						chainSet.Register(fmt.Sprintf("%s-ping", uHash), targetMachineCalc[uHash].Ping)
 						chainSet.Register(fmt.Sprintf("%s-jitter", uHash), targetMachineCalc[uHash].Jitter)
 						chainSet.Register(fmt.Sprintf("%s-fping", uHash), targetMachineCalc[uHash].FPing)
 					}
+					// finally, adds the route to the matrix
```
Harkishen-Singh, line 141:
```suggestion
					// Finally, add the route to the matrix (in-memory representation of targets).
```

Harkishen-Singh (changes requested):
Few more changes required and then we are good to go. @aquibbaig

src/lib/api/api.go
aquibbaig: Do you think we can drop the `done` channel?

Harkishen-Singh (approved):
LGTM except for the lint issue.

aquibbaig:
Will solve that and merge, thanks

PR #335, opened 2021-03-11
Fix port values in makefile and comments

dev117uday:
The pull request contains

- Fixes to all mentions of port 9090
- No need for `sudo` when running the `yarn start`

Also, all the endpoints in the postman collection makes the request to port :9090, that also needs to be updated


Harkishen-Singh (dismissed):
Thank you for this.

aquibbaig (approved):
LGTM

dev117uday:
In order to solve #336, disabling `journalctl` isn't a viable option just because the docker container runs on alpine OS. Just disabling it wont work and would further break other pieces of code. A better solution is to replace the base OS of the docker container to with something that supports `systemd`. centOS was the perfect fit for it.

As far the the size of container is concerned, changing to centos didn't make much of a difference, the size is 420Mb larger, but a much better solution to the problem.

docker
- `bench-route` is the current
- `bench-routev2` is the container with centos as base

Harkishen-Singh:
Increasing the image size by 0.5 GB is never a good thing. We can go for disabling since journal isn't much of implementation/use at the moment. (it just counts the events which are fine if we ignore them)

Harkishen-Singh:
@dev117uday can you do the `fixing dockerfile` work in a different PR and revert it from here?

muskankhedia:
@dev117uday Can you please change the port values in the file, `/tests/bench-routes-socket-tester.html` and `tests/browser.js`. This might not be required in the long run but will be really helpful to maintain consistency through out the code base.

Harkishen-Singh:
@dev117uday you have conflicts. Would you mind resolving them?

dev117uday:
Done

Harkishen-Singh (approved):
Thank you for the PR.

Harkishen-Singh:
@dev117uday please avoid making PRs from your master branch. This will block you from making more PRs until the current ones get merged.

PR #339, opened 2021-03-13
Remove dashboard v1.0

muskankhedia:
Currently, the code supported two versions of the dashboard which created confusion for many contributors. But we don't use the dashboard v1.0 anymore, I have removed all the code related to it.


Harkishen-Singh (approved):
Finally, we are getting rid of `v1.0`. :v:

Issue #257, opened 2020-06-06
UI for routes input for monitoring


Harkishen-Singh:
Provide UI for taking routes as input for monitoring screen.

Harkishen-Singh:
Right now, the input section is in the sidebar. This was done was debugging purposes. We should remove it from the sidebar to a floating-button in the monitoring screen.

Anuragtech02:
Hey @Harkishen-Singh I'd like to give this a try! Could you please assign this to me ?

Harkishen-Singh:
@Anuragtech02 please go ahead. Assigning wouldn't be the best thing for us but you can always voluntarily comment based on your interest and let others know.

Anuragtech02:
> @Anuragtech02 please go ahead. Assigning wouldn't be the best thing for us but you can always voluntarily comment based on your interest and let others know.
Thanks @Harkishen-Singh ! I'll start working on this.

PR #340, opened 2021-03-13
Moved quick-input from sidebar to FAB in monitoring Solves #257

Anuragtech02:
### What was the issue ?
Right now, the input section is in the sidebar. This was done for debugging purposes. We should remove it from the sidebar to a floating-button in the monitoring screen.

### What does this PR do ?
Solves #257 
- [x]  Creates a component called `<InputFab />` inside the file `Monitoring` itself. The component is a wrapper of material-ui's Fab and Tooltip.
 - [x]  Removes `quick-input` listItem from sidebar.

### DIFF
--- a/dashboard/v1.1/src/pages/Monitoring/Monitoring.tsx
+++ b/dashboard/v1.1/src/pages/Monitoring/Monitoring.tsx
@@ -4,11 +4,9 @@ import { HOST_IP } from '../../utils/types';
 import Matrix from './Matrix';
 import { TimeSeriesPath, RouteDetails } from '../../utils/queryTypes';
 import RouteDetailsComponent from './RouteDetails';
-import { Card, CardContent, Tooltip, Fab, makeStyles } from '@material-ui/core';
+import { Card, CardContent, Tooltip } from '@material-ui/core';
 import Alert from '@material-ui/lab/Alert';
 import Switch from '@material-ui/core/Switch';
-import { PostAdd as PostAddIcon } from '@material-ui/icons';
-import { Link } from 'react-router-dom';
 
 interface MonitoringProps {
   updateLoader(status: boolean): void;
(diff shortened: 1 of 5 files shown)


muskankhedia:
@Anuragtech02 Please add some screenshots or videos of the working of the added feature. This helps us to review the PR in an effective way and also stays as a document in the long run.

dashboard/v1.1/src/pages/Monitoring/Monitoring.tsx
```diff
@@ -80,54 +82,89 @@ const Monitoring: FC<MonitoringProps> = ({ updateLoader }) => {
...
+        </CardContent>
+      </Card>
+      <InputFab />
+    </>
   );
 };
 
 export default Monitoring;
+
+const useStyles = makeStyles(theme => ({
```
muskankhedia, line 146: Can you create a new file with the `InputFab` component?

muskankhedia:
Also, I feel we can add the `InputFab` component to all the pages including `Config`, `Ping`, `Jitter`, etc. This will be more interactive for the user, as navigating into the `monitoring` screen every time for creating a new input doesn't seem relevant. 

@Harkishen-Singh what are your views in it?

dashboard/v1.1/src/pages/Monitoring/Monitoring.tsx
Anuragtech02: Sure @muskankhedia, I'll create a separate file. Also, should I delete the listItem from sidebar because as of now I've just commented it out ?

muskankhedia: Yes, you should delete the list item of `quick input` from the sidebar.

Harkishen-Singh:
@muskankhedia, I feel if we want to have more pages than monitoring (which is fine), then the only thing I am concerned about is, that the button should be much descriptive that it adds for monitoring and not something on that page. Maybe do that as a tip if possible.

muskankhedia:
Yes we can add a tooltip in that case.

Anuragtech02:
@muskankhedia Please let me know if I have to make the above discussed changes (add fab on each page).

muskankhedia:
@Anuragtech02 Yes, you should add `fab` on every page, along with the tooltip describing the purpose of the icon to avoid confusion among the users.

Anuragtech02:
> @Anuragtech02 Yes, you should add `fab` on every page, along with the tooltip describing the purpose of the icon to avoid confusion among the users.

@muskankhedia  Could you please share what I have to add as title in the `tooltip` ? Right now I have set it to `Quick Input`

muskankhedia:
`Quick Route Input` sounds good. Can you please explain the meaning of `InputFab` here? Does the name have any specific significance?

Anuragtech02:
> `Quick Route Input` sounds good. Can you please explain the meaning of `InputFab` here? Does the name have any specific significance?

`InputFab` here means Quick Input Floating Action Button

Anuragtech02:
Hey @muskankhedia @Harkishen-Singh , I've created a separate file named `QuickInputFab` and added the fab to all pages.

Anuragtech02:
> Hey @muskankhedia @Harkishen-Singh , I've created a separate file named `QuickInputFab` and added the fab to all pages.

muskankhedia (approved):
LGTM!

PR #355, opened 2021-03-14
WIP: Improved TSDB v2

Harkishen-Singh:
This is a draft PR. Complete details will be written once things are implemented.

Silent features (mostly TODO):
1. Columnar based inserts
2. ability to handle unordered inserts
3. binary tsdb
4. indexing for faster compression
5. zlib (or lzw) based compression
6. Parallel writers for improved throughput.
7. Parallel readers
8. Singular file (against earlier file per time-series)


dbv2/reader.go
```diff
@@ -0,0 +1,189 @@
...
+	"io"
+	"os"
+	"strconv"
+)
+
+var (
+	readerBufferSize = validLength * (bufferSize / 10)
+	// readerOffsetBytes represents the number of bytes that occupy the meta-data section of the table. This can be discarded
+	// during parse operation as this is only to know the type of the table.
+	readerOffsetBytes = len([]byte(`maxValidLength: 100 chars
```
aquibbaig, line 16: What does the table metadata comprise? The hash value of routes?

dbv2/reader.go
```diff
@@ -0,0 +1,189 @@
...
+		},
+		reader: bufio.NewReaderSize(f, readerBufferSize),
+	}, nil
+}
+
+// discardInitialOffset discards the initial meta-data of the table file.
+func (dr *DataReader) discardInitialOffset() error {
+	discarded, err := dr.reader.Discard(readerOffsetBytes)
+	if err != nil {
+		return fmt.Errorf("discard initial offset: %w", err)
```
aquibbaig, line 47: Need a way to panic here and send a 500 response don't you think?

dbv2/reader.go
Harkishen-Singh: I will write a design doc for that. Mentioning it here will be very confusing to understand.

dbv2/reader.go
Harkishen-Singh: No, we should never panic. That will stop the flow of Bench-routes. Systems should be recoverable. Panicking makes them fatal and hence non-recoverable. Errors should always be traced.

dbv2/reader.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+	tr.rowsItr = newRowsIterator(tr.reader, numBytesSingleLine)
+	return tr, nil
+}
+
+// discardInitialOffset discards the initial meta-data of the table file.
+func (dr *TableReader) discardInitialOffset() error {
+	discarded, err := dr.reader.Discard(readerOffsetBytes)
+	if err != nil {
+		return fmt.Errorf("discard initial offset: %w", err)
```
chetanyakan, line 51:
```suggestion
		return fmt.Errorf("failed to discard initial offset: %w", err)
```

devansh42:
Hey @Harkishen-Singh, I was whiling around the code, and have a doubt regarding the Reader.
We are using *bufio.Reader* for reading and *bufio.Writer* for writting as it minimizes disk writes, bufio.Reader does the same for disk reads. But in tableRowIterator, We are skipping rows by reading them, Isn't it inefficient?
Can't we use File Seeking to jump to the right place and then use a bufio.Reader to read the whole chunk at once or in fewer Reads, but it requires a buffer allocation for every new bufio.Reader and We can implement our own buffer reader, which reuses the given buffer.

Harkishen-Singh:
Hey @devansh42. Yes you are right, the aim behind bufio is to minimise writes so that in high throughput scenario, tsdb continues to perform well.

We do read all rows except when being called inside reader.Parse(), since the aim of parse is to get to the last line and send the max timestamp of the file. When we are just concerned with max timestamp in a file, then parsing all the rows (other than the last one) does not have any use case, rather we waste the cpu.

The reason we need to get the max timestamp is, when the db starts, rows are maintained in increasing order of time. This means all I coming rows into the file needs to be sorted and greater than that last row (which has maxt). Otherwise we have to shift the rows down to accomodate another row.

In order to make sure all rows are in increasing order of timestamp, we store the rows in tableBuffer (in memory, hence able to handle high cardinality) and once the buffer gets loaded, we sort and flush and Increment the permitted maxt.

Further, we do implement direct jumps to the line location which contains the interested series_id. See the ReadAll() func, it takes the indexes array, and then does jumps by skipping those bytes than reading, which forms a sorted dense index and querying is faster. Note that we skip all lines only when db starts, which happens only in Parse() func, as the aim/reason is said above.

When querying will happen, the func executing the query will send the indexes array along with mint and maxt of query. Hence, there wil be a Read() func that takes those 3 params (indexes, mint, maxt) and returns the rows. This will be an extension to ReadAll()

I hope this gives some insight on what's happening. I am super happy to see your interest in this. Happy to accept more suggestions and answer any queries.

Looking forward for more collaboration.

devansh42:
Hey @Harkishen-Singh, Thanks for such a descriptive explanation. I found some doubts to discuss.

1. bufio.Reader.Discard guarantees not to read underlying io.Reader only if bytes to discard are less than the number of bytes not read from the internal buffer yet. 
2.  If we closely look at the implementation of the Discard method, it just plays with indices of its internal buffer, which means data is already read from the underlying io.Reader but is skipped to be explicitly read by bufio.Reader. Also, the Discard method uses a method *fill* which reads from the underlying io.Reader. So Anyhow we are unnecessarily reading data.
3. In the *Parse* method, We are parsing the whole file just to read the timestamp of the last row, Can't we just add a metadata row at the beginning of the file, So we don't need to read the whole file.

P.S. Please correct me if I'm wrong somewhere with bufio.Reader.Discard

Issue #295, opened 2020-08-30
Measuring query times from the TSDB


aquibbaig:
It is good to have a metric which measures `query_time` of a particular query from our Querier for a datapoint stored in TSDB. This will not only help in performance analysis, it will also help in benchmarking our architecture for good.

Harkishen-Singh:
This should also be shown in the UI (like `returned in ...sec`)

Tushar3099:
I want to work on this issue @Harkishen-Singh.

Harkishen-Singh:
Please go ahead @Tushar3099

Tushar3099:
IMHO I think inside `querier.go`, we have to modify the `exec` function and `Query` struct. We have to calculate the time spent on query by : 
`start:= time.Now()`
`// Querying the query`
`execTime:= time.Since(start)`
Then add the value of execTime to Query(which will have an extra field of execTime).
Guide me on what will be the right approach? @Harkishen-Singh

Harkishen-Singh:
@Tushar3099 that is indeed correct.

Tushar3099:
@Harkishen-Singh As I saw code, in `exec` function the `queryTime` is already implemented. The QueryResponse has a field of `TimeInvolved` which calculates the query execution time. Also in the frontend, the queryTime is sent when we hit the `/query-matrix` endpoint.
Where do we want to show this queryTime exactly?

Harkishen-Singh:
I think the ideal way would be to show in top right corner as queried in .... seconds. Above that, we should also show a time for fetching data and this time will be calculated (from start of request till receiving the data) in UI. 

In short, the time for fetch = time of querying + network time.

PR #364, opened 2021-03-15
Add Querier support for Graphs

ruddi10:
Signed-off-by: ruddi <rudrakshaggarwalsachin@gmail.com>

fixes #258 
fixes #383 
### **Feature Added-**
Querier Support for  graphs.

Peek 2021-03-15 22-56

PR #376, opened 2021-03-17
Adding FetchTime and QueryTime for each Query

Tushar3099:
Fixes Issue: #295 

**Type Of Change**
----

- [x] New feature (non-breaking change which adds functionality)

**Changes**
----

- Modified  `QueryResponse` to send queryTime in Nanoseconds.
- Refactored some functions in `Matrix.tsx`
- Calculated fetchTime for fetching query.
- Modifying HTML structure and styling for showing URL.
- Added fetchTime and queryTime in the top-right corner.

**Preview**
----

https://user-images.githubusercontent.com/51849085/111930492-16376000-8adf-11eb-84af-530bc0aa3052.mp4


dashboard/v1.1/src/pages/Monitoring/RouteDetails.tsx
```diff
@@ -115,23 +122,49 @@ const RouteDetailsComponent: FC<RouteDetailsProps> = ({
...
+              padding: '0.4rem 0.4rem 0 0.4rem',
+              margin: 0,
+              maxWidth: '90%',
+              // width: '200px',
+              overflow: 'hidden',
+              textOverflow: 'ellipsis',
+              whiteSpace: 'nowrap'
+            }}
+          >
+            {/* {truncate(data.name, 70)} */}
```
aquibbaig, line 153: Is this PR ready for review? Please request reviews once it is and remove comments before doing that

Tushar3099: Thanks for the reply @aquibbaig, will do the prescribed changes and request for review.

Tushar3099:
@Harkishen-Singh @aquibbaig, my PR is ready for review. Please review it.

tsdb/querier/querier.go
```diff
@@ -134,7 +134,7 @@ func (q *Query) exec(blockStream []tsdb.Block, encoding bool) interface{} {
 		// A nil range represents to return all time-series value as response.
 		if q.Range == nil {
 			base = QueryResponse{
-				TimeInvolved: time.Since(stamp),
+				TimeInvolved: time.Since(stamp).Nanoseconds(),
```
Harkishen-Singh, line 137: I wouldn't do this here. The way this should be done is, at line 103, we must use a defer so that the only statement works for all cases and we make sure that there are no edge cases.

```golang
defer func() {
	q.queryResponse.TimeInvolved = time.Since(stamp).Nanoseconds()
}()
```

Harkishen-Singh: And remove all other places for this calculation, as this itself is enough.

Note: Query time is the time from when you enter this func to exit of this func, since that's where the actual querying takes place. All other times fall under fetch time.

dashboard/v1.1/src/pages/Monitoring/RouteDetails.tsx
```diff
@@ -115,23 +121,47 @@ const RouteDetailsComponent: FC<RouteDetailsProps> = ({
...
-      </span>
+        <div>
+          <h6 style={{ margin: 0, fontWeight: 'bold' }}>
+            Fetch Time {'    '}: {data.fetchTime.toFixed(3)} ms
+          </h6>
+          {value && data.queryTime[value - 1] ? (
+            <h6 style={{ margin: 0, fontWeight: 'bold' }}>
+              Query Time {'  '}: {data.queryTime[value - 1] / 1000} &micro;s
+            </h6>
+          ) : null}
```
Harkishen-Singh, line 162: 2 points:
1. What happens if the time is never in ms or micros? If it's in seconds, then I guess this will show some thousand ms which is never good.
2. I will be happy if this is move concise. A bold query-time is never good. Let's make it compact and cover the timing with a beautify badge that can be green in colour.

dashboard/v1.1/src/pages/Monitoring/RouteDetails.tsx
```diff
@@ -55,6 +54,8 @@ const format = (data: RouteDetails) => {
   const pingMax: chartData[] = [];
   const jitter: chartData[] = [];
   const name = data.name;
+  const fetchTime = data.fetchTime;
+  const queryTime: number[] = [0, 0, 0];
```
Harkishen-Singh, line 58: Can we not have this as an object? using an array for this purpose is weird.

dashboard/v1.1/src/utils/queryTypes.ts
```diff
@@ -116,4 +116,5 @@ export interface RouteDetails {
   jitter: QueryResponse;
   monitor: QueryResponse;
   name: string;
+  fetchTime: number;
```
Harkishen-Singh, line 119:
```suggestion
  evaluationTime: number;
```

tsdb/querier/querier.go
Harkishen-Singh: In order to save from computations in UI, How about using `time.Since(stamp).String()`?

Harkishen-Singh (changes requested):
Some questions:
1. I need to double-check, but I think we just query once a metric. That means, the query & fetch timings should never change even when the section is changed.

Please let me know if I am wrong.

Harkishen-Singh:
@aquibbaig can you please give a quick review of the react code?

Tushar3099:
> Some questions:
> 
> 1. I need to double-check, but I think we just query once a metric. That means, the query & fetch timings should never change even when the section is changed.
> 
> I am happy to be wrong. Please point me out.

@Harkishen-Singh Actually inside SendMatrix function, I observed that : 
```
		go parallelQueryExec(matrix.PingChain.Path, querier.TypeFirst, curr, from, chans[0])
		go parallelQueryExec(matrix.JitterChain.Path, querier.TypeFirst, curr, from, chans[1])
		go parallelQueryExec(matrix.MonitorChain.Path, querier.TypeFirst, curr, from, chans[2])
		matrixResponse = map[string]querier.QueryResponse{
			"ping":    <-chans[0],
			"jitter":  <-chans[1],
			"monitor": <-chans[2],
		}
```
each of the queries was done parallelly in the different channels that took different time durations to complete so I assumed it will be different for each section. Please correct me if I misunderstood something.
Also, I want to ask about the `fetchTime` implementation that I did in the frontend. I actually calculated the time duration for the API to return data and assigned it to `fetchTime`. Is that implementation correct?

tsdb/querier/querier.go
Tushar3099: - Absolutely right, `defer` will be best for the time duration calculation here. I will implement it that way.
- Yes, the `time.Since(stamp).String()` will help us avoid the calculations in the frontend, also then the return type will be a string I assume. I had a question, sending time as a string will prevent us from doing any calculation on the time in the frontend, that shouldn't be a problem, right?

dashboard/v1.1/src/pages/Monitoring/RouteDetails.tsx
Tushar3099: - I can use the moment library to handle the time symbols. That should solve the issue. May I use that?
- I will surely do the prescribed styling changes.

dashboard/v1.1/src/pages/Monitoring/RouteDetails.tsx
Tushar3099: My bad, an object struct will be a better fit for the use case. I will do the changes.

tsdb/querier/querier.go
Harkishen-Singh: I don't think there are any calculations involved with time in the frontend, so this should be fine.

dashboard/v1.1/src/pages/Monitoring/RouteDetails.tsx
Harkishen-Singh: point 1 will be fixed if we follow https://github.com/bench-routes/bench-routes/pull/376#discussion_r598098250.

Harkishen-Singh:
> queries was done parallelly in the different channels that took different time durations to complete so I assumed it will be different for each section.

Ah, I missed that point. That makes the implementation correct in monitoring screen.

BTW, I don't see the code for query and fetch timings in all other graphs. Are they still under WIP? @Tushar3099 

PS: there are compilation errors that needs a fix.

dashboard/v1.1/src/pages/Monitoring/RouteDetails.tsx
Tushar3099: Yeah, using `time.String()` will fix the evaluation time automatically. I was asking about the fetch time, which I am calculating in the frontend because that will give me result in milliseconds : 
```
      const start = performance.now();
      const details = await fetchDetails(instance);
      const end = performance.now();
```

Tushar3099:
Yes, @Harkishen-Singh the last commit was done by mistake, right now I am implementing all the prescribed changes.
Do I have to show the query and fetch time in any other place also rather than the Routes Detail Screen? Please tell me if I overlooked something.

dashboard/v1.1/src/pages/Monitoring/RouteDetails.tsx
Harkishen-Singh: Yeah, you should make sure that all units are handled

Harkishen-Singh:
Query and fetch times would be shown in all screens that have a graph which are being fetched through the API.

For example, dashboard has 7 graphs in total, but we should have only 2, the first for syste metrics and second for jounrla metrics. The test screen will also have a similar story (single fetch & query time on each dropdown click).

@Tushar3099

dashboard/v1.1/src/pages/Monitoring/Matrix.tsx
```diff
@@ -51,50 +51,58 @@ const Element: FC<ElementProps> = ({
...
+  };
+  const fetchTimeSeriesDetails = async (instance: TimeSeriesPath) => {
+    try {
+      const start = performance.now();
+      const details = await fetchDetails(instance);
+      const end = performance.now();
+      showRouteDetails(true, {
+        ...details,
+        name: instance.name,
+        fetchTime: end - start
```
aquibbaig, line 77: `fetchtime` is not a prop of type `RouteDetails`.

dashboard/v1.1/src/pages/Monitoring/Matrix.tsx
```diff
@@ -51,50 +51,58 @@ const Element: FC<ElementProps> = ({
...
-      fetchDetails();
-    });
-
-    const details = await monitoringDetails;
-    showRouteDetails(true, { ...details, name: instance.name });
-  };
-
-  useEffect(() => {
-    async function fetchMatrix(name: string) {
+  const fetchDetails = async (instance: TimeSeriesPath) => {
```
aquibbaig, line 54: NIT: Let's add the return type (in other places too)
```diff
- const fetchDetails = async (instance: TimeSeriesPath)
+ const fetchDetails = async (instance: TimeSeriesPath): Promise<RouteDetails>
```

dashboard/v1.1/src/pages/Monitoring/RouteDetails.tsx
aquibbaig: Please add a comment on why `queryTime` has a structure as you have mentioned (e.g. ping, jitter, etc.)

aquibbaig (changes requested):
Great work! This PR is looking good, with small changes here and there. From a design perspective, will it be better to show something like "query returned in x ms"? That would depict our actions better as providing positive user feedback.

Tushar3099:
@Harkishen-Singh @aquibbaig  I am making all the prescribed changes currently and will make the commit ASAP.

Tushar3099:
@Harkishen-Singh @aquibbaig  I tried to implement all the prescribed changes up to my best. Right now, `fetchTime` and `queryTime` are only shown in the Monitored Routes screen. Showing them on the Dashboard screen and Test screen is still WIP.

I have updated the Preview to show the current changes.

Harkishen-Singh:
Short nit: when we initially click on the route, and the response length tab is first which is opened, why don't the query time gets shown there? I see only the fetch time being shown.

Tushar3099:
@Harkishen-Singh Actually we are getting the queryTime as : 
```
              {
                    ping : "string",
                    jitter : "string",
                    monitor : "string"
              }
```
So I am showing the respective times in respective sections : 
```
            queryTime(ping) --> Ping Section
            queryTime(jitter) --> Jitter Section
            queryTime(monitor) --> Response Delay Section
```
I was not sure which `queryTime` to show in the ResponseLength Section so I omitted queryTime there. Please pardon my mistake and guide what will be the right thing to do.

tsdb/querier/querier.go
```diff
@@ -95,11 +98,26 @@ func (q *Query) ExecWithoutEncode() QueryResponse {
...
 	base, stamp := getBaseResponse(q.Range)
+	defer func() {
+		var res QueryResponse
+		if q.encoding {
+			err := json.Unmarshal(i.([]byte), &res)
+			if err != nil {
+				log.Errorln(fmt.Errorf("Decoding error: %s", err.Error()).Error())
+			}
+			res.EvaluationTime = time.Since(stamp).String()
+			i, _ = json.Marshal(res)
```
Harkishen-Singh, line 114:
```suggestion
			if i, ok = json.Marshal(res); !ok {
			    log.Error("invalid type received")
			}
```

we should log error if not ok is false

Harkishen-Singh (approved):
A small change would have been better, but considering this will be updated in binary tsdb, I wouldn't block this. Thank you very much.

Harkishen-Singh:
@Tushar3099 can you please address my last comment and resolve the commit?

tsdb/querier/querier.go
```diff
@@ -95,11 +98,26 @@ func (q *Query) ExecWithoutEncode() QueryResponse {
...
+		} else {
+			res, _ = i.(QueryResponse)
+			res.EvaluationTime = time.Since(stamp).String()
+			i = res
+		}
+	}()
```
Harkishen-Singh, line 120: This can be written as.

```suggestion
	defer func() {
		var res QueryResponse
		if !q.encoding {
		        res, ok := i.(QueryResponse)
		        if !ok {
		           log.Error("type 'QueryResponse' not received")
		        }
			res.EvaluationTime = time.Since(stamp).String()
			i = res
			return
		}
		err := json.Unmarshal(i.([]byte), &res)
		if err != nil {
			log.Errorln(fmt.Errorf("Decoding error: %s", err.Error()).Error())
		}
		res.EvaluationTime = time.Since(stamp).String()
		i, ok = json.Marshal(res)
		if !ok {
		      log.Error("invalid marshalling: ", err.Error())
	        } 
	}()
```

Harkishen-Singh:
Let me know when this is ready for review

Tushar3099:
Yeah, I will update soon, actually, there are some bugs with the recent changes. I am resolving them and will update the PR asap.

Tushar3099:
@Harkishen-Singh, there are many conflicts between this PR and #364, as both have the same working files, many changes are overwriting each other, so it will take some time to resolve and update. Will try to do it asap.

Tushar3099:
@Harkishen-Singh @aquibbaig I have merged my PR with #364. Please review it.

Harkishen-Singh (approved):
LGTM! Thank you for this. However, this will have to wait before #439 because of some performance issues. Sorry for the inconvenience. I will merge once that PR is merged.

Tushar3099:
@Harkishen-Singh, I can revert to the last commit so that it won't have any dependency on #439, and then we can merge it normally because my PR has no dependency in itself with #439.

Harkishen-Singh:
Let's wait for a day or two. We are testing the that commit. If it seems fine, we won't revert then. So better to hold on at present.

Tushar3099:
@Harkishen-Singh @aquibbaig Please review and merge my PR as the revert of PR #439 is closed, for which my PR was put on hold.

Tushar3099:
@Harkishen-Singh Can you please review and merge this PR? Thank you

Harkishen-Singh (approved):
Thank you for this. Just a note to myself, in the new UI, the timings can be made more compact.

PR #380, opened 2021-03-17
Improved labelselector

### LINKED ISSUE #365: Add Label option in Quick Input
**Is your feature request related to a problem? Please describe.**
When we create a label we have to again select that label and apply that.

**Describe the solution you'd like**
As soon as we create a label it should apply.

codetheorem:
Signed-off-by: codetheorem <hrishikeshagarwalv@gmail.com>

fixes #365 

Improved functioning of  Label selector in quick input page.
Removed done icon as it  both done icon and close icon were appearing simultaneoulsy, selected tab can be seen via color change and close button appearing so done icon is not required in this case.

### DIFF
--- a/dashboard/v1.1/src/pages/Input/components/MaterialLabelSelector.tsx
+++ b/dashboard/v1.1/src/pages/Input/components/MaterialLabelSelector.tsx
@@ -2,7 +2,7 @@ import React, { Suspense, useEffect, useState } from 'react';
 import { useTheme } from '@material-ui/core/styles';
 import {
   HelpOutline as HelpOutlineIcon,
-  Close as CloseIcon,
+  Done as DoneIcon,
 } from '@material-ui/icons';
 import Autocomplete, {
   AutocompleteCloseReason
@@ -63,7 +63,7 @@ export default function MaterialLabelSelector(props: LabelSelectorType) {
         return labels;
       })
       .then(labels => {
-        if(labelList.length===0)
+        if(labelList.length===0) // This ensures that labelist from backend is only updated once at first rendering of the page.
         setLabelList(labels);
         setValue(defaultLabels);
         setPendingValue(defaultLabels);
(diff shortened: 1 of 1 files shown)


dashboard/v1.1/src/pages/Input/components/MaterialLabelSelector.tsx
```diff
@@ -43,6 +42,9 @@ export default function MaterialLabelSelector(props: LabelSelectorType) {
       name: labelValue,
       color: stc(labelValue)
     };
+    setPendingValue([...pendingValue, newLabel]);
+    setValue(pendingValue);
+    updateLabels(pendingValue);
```
aquibbaig, line 47: Can you describe in detail how this change improves the LabelSelector in the PR description, would be useful for folks to keep that in mind while reviewing this PR.

aquibbaig (commented):
This PR looks good to me, but keep in mind that we will completely let go of "labels" in v2-dashboard (which is a GSoC idea as in now)

dashboard/v1.1/src/pages/Input/components/MaterialLabelSelector.tsx
codetheorem: @aquibbaig I have added functionality that as soon as you create a new label it is applied automatically like in GitHub when we create an issue label it is applied on its own. Plus there was a bug that once you select a created label you cannot undo it but now if you want to remove that label you can do that and still, it will remain there in the label list. In the above code label list, labels applied are being updated as soon as a new label is created.

aquibbaig (approved):
Looks good, thanks!

aquibbaig:
I think it's better to write our own "LabelSelector" in the future, maybe we can think about it this summer.

dashboard/v1.1/src/pages/Input/components/MaterialLabelSelector.tsx
```diff
@@ -61,12 +63,13 @@ export default function MaterialLabelSelector(props: LabelSelectorType) {
         return labels;
       })
       .then(labels => {
+        if(labelList.length===0)
```
aquibbaig, line 66: Might as well make this piece informative by either adding curly braces or comment on top.

aquibbaig (changes requested):
When I add a new label, it does add it to selected labels, however, it does not add it to the list of labels (labelList). I'd say let's not add it to the backend but at least update the LabelList's state on the frontend

codetheorem:
> When I add a new label, it does add it to selected labels, however, it does not add it to the list of labels (labelList). I'd say let's not add it to the backend but at least update the LabelList's state on the frontend

@aquibbaig current code removes this issue. you can check this code.

Issue #386, opened 2021-03-18
Graphs/Charts in dark mode.


rishav4101:
**Is your feature request related to a problem? Please describe.**
Graphs/charts still remain in light mode when the webpage is changed to dark mode.

**Describe the solution you'd like**
I would like to change the graphs/charts to dark mode too, with making use of 'dark-mode' variable stored in localstorage.

**Describe alternatives you've considered**
If the use of localstorage is not preferred we could pass the darkMode prop to children components from BaseLayout.tsx, which can used to change themes of those components. I have used this method in PR #357. 

**Additional context**
Adding this feature shall deem the PR #357 void, as 'theme' option to Apexcharts caters to both the needs.

**Screenshot of proposed feature**
![darkgraph](https://user-images.githubusercontent.com/55195287/111656268-6046ea00-8830-11eb-946d-11a3d3733a60.png)
![ano](https://user-images.githubusercontent.com/55195287/111678448-4e704180-8846-11eb-9f46-47d0b5817e08.png)

rishav4101:
@Harkishen-Singh @muskankhedia I would like to work on this feature.

ruddi10:
@rishav4101 I have implemented dark-mode variable in one of the pr and its merged so yes I think you can use that

ruddi10:
But I feel that here one cant see the contrast and different shades that are necessary to interpret the graphs so I think you can also think of some better contrast

rishav4101:
@ruddi10 Yes I have used localstorage for now. Passing props is just an alternative. 
About contrast, I think that can be adjusted. I have attached another Screenshot where there are different colors in the graph and it's looking fine in my opinion, please have another look at it. The slight modifications if needed can be done by modifying the colors of the graph.

muskankhedia:
> @Harkishen-Singh @muskankhedia I would like to work on this feature.

@rishav4101 Sure. Please go ahead.

PR #389, opened 2021-03-19
Added: dark-mode to Graphs/Charts

rishav4101:
Description : Graphs and Charts in dark-mode.
Signed off by : @rishav4101 rajkumarrishav4101@gmail.com
Fixes : #386 
Proof :
ano
darkgraph

### DIFF
--- a/dashboard/v1.1/src/pages/Dashboard/CPUUsage.tsx
+++ b/dashboard/v1.1/src/pages/Dashboard/CPUUsage.tsx
@@ -1,12 +1,14 @@
-import React, { FC } from 'react';
+import React, { FC, useContext } from 'react';
 import Chart from 'react-apexcharts';
 import { chartData } from '../../utils/queryTypes';
+import { ThemeContext } from '../../layouts/BaseLayout';
 
 interface CPUUsageProps {
   cpuMetrics: chartData[];
 }
 
 const CPUUsage: FC<CPUUsageProps> = ({ cpuMetrics }) => {
+  const themeMode = useContext(ThemeContext);
   const dataFormatted = cpuMetrics;
   const series = [
     {
@@ -27,8 +29,7 @@ const CPUUsage: FC<CPUUsageProps> = ({ cpuMetrics }) => {
           hideOverlappingLabels: true,
           trim: true
         }
-      },
-      background: '#fff'
+      }
     },
     dataLabels: {
       enabled: false
@@ -54,6 +55,9 @@ const CPUUsage: FC<CPUUsageProps> = ({ cpuMetrics }) => {
         opacityFrom: 1,
         opacityTo: 0.2
       }
+    },
+    theme: {
+      mode: themeMode
     }
   };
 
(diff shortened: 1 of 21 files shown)


rishav4101:
@Harkishen-Singh @muskankhedia Please review it.

dashboard/v1.1/src/pages/Dashboard/CPUUsage.tsx
```diff
@@ -54,6 +53,9 @@ const CPUUsage: FC<CPUUsageProps> = ({ cpuMetrics }) => {
         opacityFrom: 1,
         opacityTo: 0.2
       }
+    },
+    theme: {
+      mode: localStorage.getItem('dark-mode') === 'true' ? 'dark' : 'light'
```
aquibbaig, line 58: Let's not reference localStorage in every case, what I'd say is to use localStorage once(if you can) in any top-level component and pass it as a context variable. The issue is that the read performance of localStorage can be costly at times. I've noticed my browser freezing on this branch and referencing localStorage constantly can be a cause. Also if all the charts have the same theme prop, what about instantiating it once and referencing it everywhere.

aquibbaig (changes requested):
Hey, thanks for this PR! I can see that my browser freezes for a couple of seconds when I toggle dark mode, (dark mode should be seamless). There are a couple of reasons why it may happen and how we can fix it.
## Multiple references of localStorage in our code
 - jsperf says that the read performance of `localStorage` is significantly less than `native`. It may have less effect on real world uses but even if it doesn't, we must look for best practices here.
 - The solution here is to gather all references of theme variable and create a context variable in any top-level component and pass this as a prop to page component where you can say `const { darkMode } = useContext(Theme)`. 
## Multiple re-renders of charts cause the page to freeze.
- Since on a dark mode, we re-render charts with all their data in memory, this can be a costly operation.
- The solution here is to use code-splitting and use `React.Suspense` and `React.lazy` to show loaders until the charts are loaded and display them after that.

muskankhedia:
@rishav4101 Can you please fix this?

Screenshot 2021-03-20 at 2 37 22 PM

rishav4101:
@muskankhedia @aquibbaig Kindly review it.

muskankhedia:
@rishav4101 Can you please squash all the commits to one commit?

muskankhedia (changes requested):
Can you please fix the graph on the `monitoring` and `tests` screen as well?

Screenshot 2021-03-22 at 11 24 51 AM

Screenshot 2021-03-22 at 11 25 55 AM

muskankhedia:
@rishav4101 If you check the `Files changed` now, you could see the previously edited files have been edited in your commit. I will advise you to create a new PR only with your changes or revert the changes in the files that aren't a part of this PR. Can you please pick any one of the option and edit accordingly?

rishav4101:
Yes, I will create a new PR for the change.

rishav4101:
I have created a new PR #420, closing this one.

PR #391, opened 2021-03-19
Added Message field to send appropriate messages from API 

### LINKED ISSUE #337: send appropriate messages from the API
Rather than throwing an error on the console on errors or just a "success" alert on success , let's send an appropriate message from the API and display it as a toast alert on the dashboard.
Currently, we send the API interface in the form of a byte array.
An approach to this problem, IMHO will be to have an interface for `response_object` and populate the same before sending it. It can contain some primitive parameters such as `Data/Body`, `Message`, etc.

Rits1272:
Signed-off-by: rits1272 ritikjain1272@gmail.com

File Changed: `api.go`

 ### Changes
 - Added `Message` field in the response to send appropriate messages from the API
 - This `message` can then be displayed as a `toast` in the dashboard   
 - The errors which were handled using `panic`, an appropriate message has been added and sent to the frontend 
 - Uptil this commit, no changes have been made in the frontend. Once get approval for the API changes, complimentary changes can be made at the frontend side either in this PR or a new PR

Resolves #337

### DIFF
--- a/src/lib/api/api.go
+++ b/src/lib/api/api.go
@@ -35,6 +35,7 @@ const (
 type API struct {
 	ResponseStatus    string
 	Data, Services    interface{}
+	Message           string
 	Matrix            *map[string]*utils.BRMatrix
 	configurationPath string
 	config            *config.Config
@@ -293,14 +294,18 @@ func (a *API) SendMatrix(w http.ResponseWriter, r *http.Request) {
 // of the react-UI.
 func (a *API) QuickTestInput(w http.ResponseWriter, r *http.Request) {
 	if err := r.ParseForm(); err != nil {
-		panic(err)
+		a.Message = err.Error()
+		a.send(w, a.marshalled())
+		return
 	}
 	var (
 		t       inputRequest
 		decoder = json.NewDecoder(r.Body)
 	)
 	if err := decoder.Decode(&t); err != nil {
-		panic(err)
+		a.Message = err.Error()
+		a.send(w, a.marshalled())
+		return
 	}
 	req := request.New(t.URL, t.Headers, t.Params, t.Body, t.Labels)
 	response := make(chan request.ResponseWrapper)
@@ -313,6 +318,7 @@ func (a *API) QuickTestInput(w http.ResponseWriter, r *http.Request) {
 		fmt.Printf("invalid request method: %s\n", t.Method)
 	}
 	a.Data = <-response
+
 	a.send(w, a.marshalled())
 }
 
@@ -323,7 +329,9 @@ func (a *API) AddRouteToMonitoring(w http.ResponseWriter, r *http.Request) {
 		decoder = json.NewDecoder(r.Body)
 	)
 	if err := decoder.Decode(&t); err != nil {
-		panic(err)
+		a.Message = err.Error()
+		a.send(w, a.marshalled())
+		return
 	}
 	requestInstance := request.New(t.URL, t.Headers, t.Params, t.Body, t.Labels)
 	a.config.AddRoute(
@@ -373,20 +381,26 @@ func (a *API) UpdateMonitoringServicesState(w http.ResponseWriter, r *http.Reque
 	service := reflect.ValueOf(a.Services).Elem()
 	sp, ok := service.FieldByName("Ping").Interface().(*ping.Ping)
 	if !ok {
-		panic("start-monitoring: ping not found")
+		a.Message = "start-monitoring: ping not found"
+		a.send(w, a.marshalled())
+		return
 	}
 
 	sp.Iterate(state, false)
 
 	sj, ok := service.FieldByName("Jitter").Interface().(*jitter.Jitter)
 	if !ok {
-		panic("start-monitoring: jitter not found")
+		a.Message = "start-monitoring: jitter not found"
+		a.send(w, a.marshalled())
+		return
 	}
 	sj.Iterate(state, false)
 
 	sm, ok := service.FieldByName("Monitor").Interface().(*monitor.Monitor)
 	if !ok {
-		panic("start-monitoring: monitor not found")
+		a.Message = "start-monitoring: monitor not found"
+		a.send(w, a.marshalled())
+		return
 	}
 	sm.Iterate(state, false)
 	a.send(w, a.marshalled())
@@ -397,21 +411,29 @@ func (a *API) GetMonitoringState(w http.ResponseWriter, _ *http.Request) {
 	service := reflect.ValueOf(a.Services).Elem()
 	sp, ok := service.FieldByName("Ping").Interface().(*ping.Ping)
 	if !ok {
-		panic("start-monitoring: ping not found")
+		a.Message = "start-monitoring: ping not found"
+		a.send(w, a.marshalled())
+		return
 	}
 
 	sj, ok := service.FieldByName("Jitter").Interface().(*jitter.Jitter)
 	if !ok {
-		panic("start-monitoring: jitter not found")
+		a.Message = "start-monitoring: jitter not found"
+		a.send(w, a.marshalled())
+		return
 	}
 
 	sm, ok := service.FieldByName("Monitor").Interface().(*monitor.Monitor)
 	if !ok {
-		panic("start-monitoring: monitor not found")
+		a.Message = "start-monitoring: monitor not found"
+		a.send(w, a.marshalled())
+		return
 	}
 
 	if sp.IsActive() != sj.IsActive() || sm.IsActive() != sp.IsActive() || sm.IsActive() != sj.IsActive() {
-		panic("states not aligned")
+		a.Message = "states not aligned"
+		a.send(w, a.marshalled())
+		return
 	}
 
 	if sp.IsActive() {
@@ -437,7 +459,9 @@ func (a *API) GetConfigRoutes(w http.ResponseWriter, _ *http.Request) {
 // ModifyIntervalDuration modifies a specific interval duration in the config file.
 func (a *API) ModifyIntervalDuration(w http.ResponseWriter, r *http.Request) {
 	if err := r.ParseForm(); err != nil {
-		panic(err)
+		a.Message = err.Error()
+		a.send(w, a.marshalled())
+		return
 	}
 	var req struct {
 		IntervalName string `json:"intervalName"`
@@ -445,7 +469,9 @@ func (a *API) ModifyIntervalDuration(w http.ResponseWriter, r *http.Request) {
 	}
 	decoder := json.NewDecoder(r.Body)
 	if err := decoder.Decode(&req); err != nil {
-		panic(err)
+		a.Message = err.Error()
+		a.send(w, a.marshalled())
+		return
 	}
 	if num, err := strconv.Atoi(req.Value); err == nil {
 		n := int64(num)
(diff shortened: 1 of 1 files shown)


Rits1272:
Kindly review @aquibbaig @Harkishen-Singh

aquibbaig:
Hey, @Rits1272 I'd like to test this PR thoroughly before reviewing it, will do it either today or tomorrow. Thanks.

src/lib/api/api.go
```diff
@@ -293,14 +294,18 @@ func (a *API) SendMatrix(w http.ResponseWriter, r *http.Request) {
 // of the react-UI.
 func (a *API) QuickTestInput(w http.ResponseWriter, r *http.Request) {
 	if err := r.ParseForm(); err != nil {
-		panic(err)
+		a.Message = err.Error()
+		a.send(w, a.marshalled())
```
Harkishen-Singh, line 298: This seems like a repeated pattern. Can we make a func call for this and send the output? This func will be called across all the functions and hence, reduce the redundancy.

Harkishen-Singh (changes requested):
Another necessary change that is not related to this PR, rather an issue in existing code which can be done in this PR:

Can we change the a.send() to accept (ResponseWriter, Data, Error, Message) as params and then marshal inside it and do the Write()?

The current code is not concurrency safe, so if more than one requests come at a time, the result will be a race condition. The above suggested change should remove this problem.

Thank you in advance for the contribution.

src/lib/api/api.go
Rits1272: Yeah sure!

Rits1272:
Hello @Harkishen-Singh, kindly review the changes :)

src/lib/api/api.go
```diff
@@ -68,6 +69,11 @@ func New(
 	}
 }
 
+func (a *API) sendMsg(msg string, w http.ResponseWriter) {
+	a.Message = msg
+	a.send(w, a.marshalled(), a.Message)
```
Harkishen-Singh, line 74:
```suggestion
	a.send(w, a.marshalled(), msg)
```

src/lib/api/api.go
```diff
@@ -35,6 +35,7 @@ const (
 type API struct {
 	ResponseStatus    string
 	Data, Services    interface{}
+	Message           string
```
Harkishen-Singh, line 38:
```suggestion
```

src/lib/api/api.go
```diff
@@ -558,16 +581,18 @@ func (a *API) GetLabels(w http.ResponseWriter, r *http.Request) {
...
-		Data   interface{} `json:"data"`
+		Status  string      `json:"status"`
+		Data    interface{} `json:"data"`
+		Message string      `json:"message"`
 	}{
-		Status: a.ResponseStatus,
-		Data:   a.Data,
+		Status:  a.ResponseStatus,
+		Data:    a.Data,
+		Message: a.Message,
```
Harkishen-Singh, line 595: please change this func to

```golang
func marshalled(status string, data interface{}) []byte {
	response := struct {
		Status string      `json:"status"`
		Data   interface{} `json:"data"`
	}{
		Status: status,
		Data:   data,
	}
	js, err := json.Marshal(response)
	if err != nil {
		fmt.Fprintln(os.Stderr, "json-marshal error: ", err)
	}

	return js
}
```
If this is stuck to a.*, then race problems will occur, Hence, make this independent.

src/lib/api/api.go
```diff
@@ -577,7 +602,8 @@ func (a *API) marshalled() []byte {
 	return js
 }
 
-func (a *API) send(w http.ResponseWriter, data []byte) {
+func (a *API) send(w http.ResponseWriter, data []byte, message string) {
+	a.marshalled()
```
Harkishen-Singh, line 606: something like this over here
```suggestion
	data := marshalled(status, data)
```

src/lib/api/api.go
```diff
@@ -577,7 +602,8 @@ func (a *API) marshalled() []byte {
 	return js
 }
 
-func (a *API) send(w http.ResponseWriter, data []byte) {
+func (a *API) send(w http.ResponseWriter, data []byte, message string) {
```
Harkishen-Singh, line 605: This will change accordingly.

Harkishen-Singh (changes requested):

Rits1272:
Kindly again check the changes @Harkishen-Singh. I sincerely appreciate such a detailed review. I am actually well versed with Python frameworks and kind of a novice in Go, so thank you for bearing with me

src/lib/api/api.go
```diff
@@ -119,7 +124,7 @@ func (a *API) Home(w http.ResponseWriter, r *http.Request) {
 	msg := "ping from " + r.RemoteAddr + ", sent pong in monitor"
 	log.Infoln(msg)
 	a.Data = msg
-	a.send(w, a.marshalled())
+	a.send(w, a.Data, "Home")
```
aquibbaig, line 127: So what's the third parameter "Home" for?

aquibbaig (changes requested):
Thanks for taking an attempt at this issue. I didn't dive into the code but can you check if it works end to end with the dashboard? I don't see the message as a part of the response.
Also, we don't want any messages from the API, but when something fails (or on success which we would like to acknowledge the user), we must return a message. On errors, it could be `404:Could not reach the jitter module`. For that, we must return a 404 on panic from API.

src/lib/api/api.go
Rits1272: Yeah, actually it is the `Message` parameter. Should I leave it as an empty string or replace it with `msg` @aquibbaig?

aquibbaig: Good question! please use nil for most of the cases, unless we need to notify the user for what exactly went wrong. That way to can conditionally check for this message parameter in "send()" or use a varadic function. I can't think of many success cases to notify the users (added route to monitoring on quick input is one of such use cases), let me know if you have sth else in mind.

Rits1272: As for now @aquibbaig, I have just added the name of the `function` of API on success.  Should I change them all to `nil`?

aquibbaig: Yes! That's right. API function as messages don't serve much of a purpose.

Rits1272:
Kindly check the changes @aquibbaig @Harkishen-Singh

src/lib/api/api.go
```diff
@@ -119,7 +124,7 @@ func (a *API) Home(w http.ResponseWriter, r *http.Request) {
 	msg := "ping from " + r.RemoteAddr + ", sent pong in monitor"
 	log.Infoln(msg)
 	a.Data = msg
-	a.send(w, a.marshalled())
+	a.send(w, a.Data, "nil")
```
aquibbaig, line 127: @Rits1272 that's not exactly what I meant.
- In Go, nil is the zero value for pointers, interfaces, maps, slices, channels and function types, representing an uninitialized value.
- nil doesn't mean some "undefined" state, it's a proper value in itself. An object in Go is nil simply if and only if it's value is nil, which it can only be if it's of one of the aforementioned types.
- An error is an interface, so nil is a valid value for one, unlike for a string. For obvious reasons a nil error represents no error.

Rits1272: Thank you for the review! I have changed the `message` from `string` to `interface`. So that we can now assign `nil` to the success message. Kindly review @aquibbaig :)

Rits1272:
Please check @aquibbaig @Harkishen-Singh

src/lib/api/api.go
```diff
@@ -558,27 +576,30 @@ func (a *API) GetLabels(w http.ResponseWriter, r *http.Request) {
...
-		Status: a.ResponseStatus,
-		Data:   a.Data,
+		Status:  status,
+		Data:    data,
+		Message: message,
 	}
 	js, err := json.Marshal(response)
 	if err != nil {
-		panic(err)
+		fmt.Fprintln(os.Stderr, "json-marshal error: ", err)
```
aquibbaig, line 594: Why do we print the error at this point?

Rits1272: Hello @aquibbaig, kindly check [this](https://github.com/bench-routes/bench-routes/pull/391#discussion_r598139953)

Harkishen-Singh: It's fine to print than to panic :smiley:

src/lib/api/api.go
```diff
@@ -119,7 +120,7 @@ func (a *API) Home(w http.ResponseWriter, r *http.Request) {
 	msg := "ping from " + r.RemoteAddr + ", sent pong in monitor"
 	log.Infoln(msg)
 	a.Data = msg
-	a.send(w, a.marshalled())
+	a.send(w, a.Data, nil)
```
Harkishen-Singh, line 123: Can you remove a.Data and simply pass Data? The dependency on a should never be there. Thats what the race is all about.

src/lib/api/api.go
aquibbaig: I'd say use "log.infoln"

Harkishen-Singh: You mean log.Error?

src/lib/api/api.go
Rits1272: Yeah sure

src/lib/api/api.go
Rits1272: what changes should I make here @aquibbaig @Harkishen-Singh?

Harkishen-Singh:
```suggestion
		log.Error("json-marshal error: ", err.Error())
```

Rits1272: Thank you!

src/lib/api/api.go
```diff
@@ -119,8 +118,7 @@ func (a *API) Register(router *mux.Router) {
 func (a *API) Home(w http.ResponseWriter, r *http.Request) {
 	msg := "ping from " + r.RemoteAddr + ", sent pong in monitor"
 	log.Infoln(msg)
-	a.Data = msg
```
Harkishen-Singh, line 121: I don't see `Data` field from *API removed yet. Could you please do that as per discussions?

src/lib/api/api.go
aquibbaig: Yep, that looks fine!

src/lib/api/api.go
Rits1272: So should I remove the `Data` field from the *API struct and make the changes accordingly everywhere?

Harkishen-Singh: Yes, that's my point from starting

Rits1272: Okay

Rits1272: Kindly review @Harkishen-Singh

src/lib/api/api.go
```diff
@@ -34,7 +34,7 @@ const (
 // API type for implementing the API interface.
 type API struct {
 	ResponseStatus    string
```
Harkishen-Singh, line 36: In the same way, responseStatus will also have race condition. Kindly change that too just like data and pass that into a.send() func. that should be the final update to this PR before lgtm.

Harkishen-Singh (changes requested):

src/lib/api/api.go
```diff
@@ -505,26 +524,27 @@ func (a *API) UpdateRoute(w http.ResponseWriter, r *http.Request) {
 			break
 		}
 	}
-	if _, err := a.config.Write(); err != nil {
```
Rits1272, line 508: I have removed this since the ResponseStatus is going to be overwritten. Please let me know if we should use `return` here inside the conditional block.

Rits1272:
Kindly review @Harkishen-Singh. Thank you

Rits1272:
Kindly give it a look @Harkishen-Singh if time permits. Thank you

PR #350, opened 2021-03-14
Fix: Remove white-dash in Tab component

iamkryptonite:
Description: Added outline style to Tabs component to remove white dashed-line after clicking.
Signed off by: @iamkryptonite thesatyakidas.00220033@gmail.com
Fixes: #349 
Proof:
ezgif com-video-to-gif

PR #395, opened 2021-03-19
Remove outline of buttons

### LINKED ISSUE #390: remove that ugly border from quick-input icon
Appears when we click on the quick-input button.

Image

*P.S. See bottom-right*

ruddi10:
Signed-off-by: ruddi <rudrakshaggarwalsachin@gmail.com>
Remove the outline from button when they are focused this problem wasnt limited to just Quick Input but also to the tab of the graph tables although #350 fixes them the code was a bit repetetive and this might occur in other components in future too so I went for a different solution. The cause of this problem was a css property applied by default on button elements by the browser.
fixes #390 
### **Solution**
Peek 2021-03-19 22-50

### **Caused due to-**
Screenshot from 2021-03-19 22-33-54

### DIFF
--- a/dashboard/v1.1/src/App.css
+++ b/dashboard/v1.1/src/App.css
@@ -7,6 +7,10 @@
   pointer-events: none;
 }
 
+button:focus {
+  outline: 0px !important;
+}
+
 @media (prefers-reduced-motion: no-preference) {
   .App-logo {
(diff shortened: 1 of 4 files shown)


ruddi10:
@aquibbaig  @muskankhedia kindly review it

muskankhedia (approved):

Issue #336, opened 2021-03-11
Docker container fails to start


dev117uday:
**Describe the bug**
After building the image, when i tried to start the container error 

```
panic: exec: "journalctl": executable file not found in $PATH
```

**To Reproduce**
Steps to reproduce the behaviour:
1. `docker build -t bench-routes . `
2. `docker run -p 9990:9990 -it bench-routes`

**Screenshots**
Output when running `docker run -p 9990:9990 -it bench-routes`
![image](https://user-images.githubusercontent.com/49728410/110816899-c7552380-82b1-11eb-800d-37c3aebd2430.png)

**System info :**
 - OS: Ubuntu 20.04
 - commit : f4356551

**Additional context**
journalctl is systemd's logging service [here](https://www.loggly.com/ultimate-guide/using-journalctl/#:~:text=Journalctl%20is%20a%20utility%20for,log%20messages%20processed%20by%20journald.) The base image used in the docker is alpine which doesnt implement systemd to init service but uses openRC [more reference, read the last line](https://en.wikipedia.org/wiki/Alpine_Linux#Features)

_Unlike many distributions, including Debian, Ubuntu, RHEL, Arch Linux and CentOS, Alpine does not use systemd._

I could be wrong about this.

dev117uday:
Also `.dockerignore` is missing in the repo, starting the build unnecessarily copy the node_modules to it.

Harkishen-Singh:
Ah, alpine does not support journalctl. It has its own story of syslog. We can make journalctl disable if its not supported on a particular platform. That would fix this issue. Would you mind doing that?

To summarise the tasks require to update current stuff:
1. Update go version in docker to `1.16.1`
2. .dockerignore addition for the above purpose (possible make a different folder with dockerfile and the .ignore file?)
3. disable journal metrics if system does not have journalctl supported

Thank you @dev117uday for reporting this.

dev117uday:
@Harkishen-Singh I would love to take up this issue

dev117uday:
@Harkishen-Singh please review the pr #335

PR #403, opened 2021-03-21
fix: Dockerfile

muskankhedia:
Fixes: #336

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,6 @@
-FROM golang:1.14.2-alpine3.11
-RUN apk add --no-cache bash nodejs npm
+FROM centos:centos8
+RUN yum -y upgrade
+RUN yum install -y net-tools nodejs npm golang
 RUN node --version
 RUN npm --version
 WORKDIR $GOPATH/src/github.com/bench-routes/bench-routes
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,5 +1,6 @@
-FROM golang:1.14.2-alpine3.11
-RUN apk add --no-cache bash nodejs npm
+FROM centos:centos8
```
Harkishen-Singh, line 1: If we really want to use non alpine versions, then shouldn't we go with fedora? It will be easier for us to test as well then. 

(I don't agree that we should change to a distro that is 10 times larger, just for journal metrics to work. IMO, we should stop journal metrics in alpine (like it does not work on Mac) and that should be the answer to the fix. Journal metrics are not that useful to worth this risk.)

Harkishen-Singh (approved):

Dockerfile
muskankhedia: Considering the `linux/amd64` OS/Arch, the compressed size for centos8 docker image is `71.7Mb` whereas for golang-alpine version is near about `100Mb`. For reference I have linked the screenshots of the details from dockerhub below:

**Golang Alpine version**
WhatsApp Image 2021-03-21 at 5 45 06 PM


**CentOS**
WhatsApp Image 2021-03-21 at 5 45 24 PM

Issue #404, opened 2021-03-21
.env file to have configPaths, ports, etc


Sanketwable:
Having environment file to store the configuration variables, ports.
For now we are having variables directly defined inside main.go file. 
eg : 
port 9990,
systemMetricsPath = "storage/system.json"
journalMetricsPath = "storage/journal.json"
configurationFilePath = "local-config.yml"
storageDir
pathFloodPing
pathMonitoring
These can be stored into .env file and can be loaded into config or parser package and can be used directly by using config package.
User need not to edit main code to change port or paths.
We can also store the endpoints of the APIs in the env.

Sanketwable:
Can I Work on this

Harkishen-Singh:
It would be better to provide this via cli flags(I would choose kingpin) as well, along with a config file. You can go ahead with this. If some value is both in the cli, and in the config file, then the priority should be given to the cli ones.

Sanketwable:
As I saw in the issue https://github.com/bench-routes/bench-routes/issues/43 we need to integrate mailing service so we will need .env to store API key or some credentials. I will create .env and will give priority to CLI flags. I will create a PR. Thanks.

Harkishen-Singh:
That will be done in gsoc period and during that implementation, the corresponding cli/config param would be added. Right now, we should just support the ones that bench-routes requires.

Sanketwable:
Ok I will create a PR for this. Thanks.

PR #417, opened 2021-03-22
Adding .env file with port and paths

Sanketwable:
I have created .env file that has port, Paths.
In case of port preference is given to argument passed while running the code.
@Harkishen-Singh Please review and let me know the changes.
Thanks
Closes https://github.com/bench-routes/bench-routes/issues/404

### DIFF
--- a/.env
+++ b/.env
@@ -1,0 +1,16 @@
+#Server Information
+PORT=9990
+
+#Metrics Path
+SYSTEM_METRICS_PATH=storage/system.json
+JOURNAL_METRICS_PATH=storage/journal.json
+
+#Config Path
+CONFIGURATION_FILE_PATH=local-config.yml
+
+#Storage paths
+STORAGE_DIR=storage
+PATH_PING=storage/ping
+PATH_JITTER=storage/jitter
+PATH_FLOOD_PING=storage/flood-ping
+PATH_MONITORING=storage/monitoring
--- a/src/lib/config/config.go
+++ b/src/lib/config/config.go
@@ -2,11 +2,24 @@ package parser
 
 import (
 	"io/ioutil"
+	"os"
 	"sync"
 
-	"gopkg.in/yaml.v2"
-
+	"github.com/joho/godotenv"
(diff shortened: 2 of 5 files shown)


src/lib/config/config.go
```diff
@@ -150,3 +163,22 @@ func GetNewRouteType(method, url string, headers []Headers, params []Params, bod
...
+		PORT = "9990"
+	}
+	SYSTEM_METRICS_PATH = os.Getenv("SYSTEM_METRICS_PATH")
+	JOURNAL_METRICS_PATH = os.Getenv("JOURNAL_METRICS_PATH")
+	CONFIGURATION_FILE_PATH = os.Getenv("CONFIGURATION_FILE_PATH")
+	STORAGE_DIR = os.Getenv("STORAGE_DIR")
+	PATH_PING = os.Getenv("PATH_PING")
+	PATH_JITTER = os.Getenv("PATH_JITTER")
+	PATH_FLOOD_PING = os.Getenv("PATH_FLOOD_PING")
+	PATH_MONITORING = os.Getenv("PATH_MONITORING")
```
chetanyakan, line 183: The variables in `src/main.go` had default values.
Should we add those default values here?

.env
```diff
@@ -0,0 +1,16 @@
+#Server Information
```
chetanyakan, line 1: Ideally, the `.env` file should be added to `.gitignore`.

You can instead add a `.env.sample` file and add instructions in the `README.md` to create this file.

chetanyakan (changes requested):

src/lib/config/config.go
Sanketwable: According to me Default value will only exists for PORT rest has to be absolute Path. So It is must to have all the paths defined.
Correct me if I'm wrong.
Thanks.

.env
Sanketwable: I thought about it. But as for now we are not having any API SECRET or KEY stored in .env so I didn't add it to .gitignore.
Thanks

Issue #413, opened 2021-03-22
Reset button in Quick Route Input is not working properly


robin828:
On clicking the reset button on (Quick Route Input page) labels are not removed.

1. Go to Dashboard
2. Click on Quick Route Input
3. Now add some data to it
4. Now press the reset button
5. see the error (label are not removed)

**Expected behavior**
Labels should also be removed by clicking reset button

**Screenshots**
![Screenshot 2021-03-22 at 6 58 28 PM](https://user-images.githubusercontent.com/61390860/111997017-9c37c300-8b40-11eb-8fd6-e23eefe7aed1.png)

There is one more problem with labels that after clicking one label we have to click somewhere to close that label dropdown.
I don't know that it is a bug or it's a feature to select multiple labels at one time. If it is a bug then I really want to work on that.

robin828:
I want to work on this issue if approved.

muskankhedia:
@robin828 Sure, you can take it up.

PR #421, opened 2021-03-23
Fixed Reset button in Quick Route Input not working properly

robin828:
#413 fixed
Please review @muskankhedia


muskankhedia (approved):
LGTM!

PR #424, opened 2021-03-23
Removing mv command and directly using -o in go build

Sanketwable:
Replacing *go build and mv main bench-routes* to *go build -o bench-routes src/main.go*


muskankhedia (approved):
Than you for the PR!

PR #426, opened 2021-03-24
quick-input issue fixed #419

### LINKED ISSUE #419: Quick Route Input should not be on there Quick Route input page.
There is no use of the Quick Route Input button in the right bottom corner on the 
Quick Route Input page because it does not have any work if we are already on this page.

Screenshot 2021-03-23 at 3 03 12 AM

robin828:
Please review @aquibbaig


ruddi10:
Thanks @robin828 for your pr can you please resolve the conflicts and also give some detail on what you are fixing or tell the issue that this pull request is solving

Issue #378, opened 2021-03-17
Querier should return data from in-memory chains that are not flushed


Harkishen-Singh:
Right now, the querier queries only the time-series files. This means that querying is possible only on the data that has been flushed and not the one in-memory. This creates a big issue. If you have `flushDuration` of 5 mins (which is the current value in default), users will have to wait for 5 mins before they get any data in the graphs. This is not good. The querier should be able to get the samples from the blocks in the chains that are in-memory and are yet to be flushed. But it is important to note that this does not lead to duplication of data/blocks when being sent to the UI.

Tushar3099:
I was also facing this issue and would like to work on this issue. @Harkishen-Singh

Harkishen-Singh:
@Tushar3099 this might be a bit tricky. Feel free to take up and ask if you need any understanding.

Tushar3099:
Thanks @Harkishen-Singh , will catch up for help when needed.

Tushar3099:
@Harkishen-Singh I went through the code and observed that : 
- On adding a route in `quick-input`, the route is added to `Cmap` of the ChainSet.
- The ChainSet is then registered.
-  A function `Run` reloads after `flushedDuration` amount of time, commits the new chains ( flushes the unflushed chains containing new blocks ).
Please guide how I should approach sending the in-memory chain to the frontend.

Harkishen-Singh:
@Tushar3099 that is correct. I think you can attempt for this.

PS: I am sorry, but it is not possible for me to write the complete approach here at the moment, but I can help with reviews and information you will need if you get stuck. You should move forward with the implementation.

Tushar3099:
@Harkishen-Singh I have made a PR for this issue. Please review it.

PR #427, opened 2021-03-25
Querying unflushed-memory-chain

Tushar3099:
Signed-off-by: Tushar3099 <tusharneogi30@gmail.com>

Fixes Issue: #378 

**Type Of Change**
----

- [x] Enhancement

**Changes**
----
- Modified `ExecWithoutEncode` function to append in-memory chain to the chain retrieved from the secondary storage.
- Modified `/query` and `/query-matrix` route handlers to access the in-memory chains and pass it to `ExecWithoutEncode`
- Modified frontend fetch routes accordingly.

**Results**
- Querying in-memory chain in `Monitoring` window and `Test` window.

**Preview**
----

https://user-images.githubusercontent.com/51849085/112421453-0d4fc400-8d55-11eb-865f-96bc8d36ddc7.mp4

### DIFF
--- a/src/lib/api/api.go
+++ b/src/lib/api/api.go
@@ -174,8 +174,10 @@ func (a *API) Query(w http.ResponseWriter, r *http.Request) {
 		startTimestamp, endTimestamp int64
 		err                          error
 	)
-	timeSeriesPath := r.URL.Query().Get("timeSeriesPath")
-
+	hash := r.URL.Query().Get("hashKey")
+	Type := r.URL.Query().Get("type")
+	var timeSeriesPath string
+	var newBlocks []tsdb.Block
 	startTimestampStr := r.URL.Query().Get("startTimestamp")
 	if startTimestampStr == "" {
 		startTimestamp = int64(math.MaxInt64)
@@ -196,17 +198,46 @@ func (a *API) Query(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
+	if hash == "" {
+		switch Type {
+		case "system":
+			timeSeriesPath = "storage/system" + tsdb.FileExtension
+		case "journal":
+			timeSeriesPath = "storage/journal" + tsdb.FileExtension
+		default:
+			a.send(w, []byte("INVALID_TYPE"))
+			return
+		}
+		if ok := tsdb.VerifyChainPathExists(timeSeriesPath); !ok {
+			a.send(w, []byte("INVALID_PATH"))
+			return
+		}
+		newBlocks = []tsdb.Block{}
+	} else {
+		if _, ok := (*a.Matrix)[hash]; !ok {
+			a.send(w, []byte("INVALID_PATH"))
+			return
+		}
+		matrix := (*a.Matrix)[hash]
+
+		switch Type {
+		case "ping":
+			newBlocks = matrix.PingChain.Chain
+		case "jitter":
+			newBlocks = matrix.JitterChain.Chain
+		default:
+			a.send(w, []byte("INVALID_TYPE"))
+			return
+		}
+
+		timeSeriesPath = "storage/" + Type + "/chunk_" + Type + "_" + hash + tsdb.FileExtension
+	}
 	// condition: only for bench-routes as per the design
 	//
 	// path should be in syntax: <DBPath>/<ofType>/chunk_<middle>_<url>.json -> non-system metric
 	// %s/system.json -> system metric
 
 	// verify if chain path exists
-	timeSeriesPath = timeSeriesPath + tsdb.FileExtension
-	if ok := tsdb.VerifyChainPathExists(timeSeriesPath); !ok {
-		a.send(w, []byte("INVALID_PATH"))
-		return
-	}
 
 	// TODO: we do not capture the block streams in memory while querying yet. They are captured only when flushed.
 	// TODO: consider cmap while querying for fetching latest blocks after shifting tsdb to binary.
@@ -214,7 +245,7 @@ func (a *API) Query(w http.ResponseWriter, r *http.Request) {
 	qry := querier.New(timeSeriesPath, "", querier.TypeRange)
 	query := qry.QueryBuilder()
 	query.SetRange(startTimestamp, endTimestamp)
-	a.Data = query.ExecWithoutEncode()
+	a.Data = query.ExecWithoutEncode(newBlocks)
 	a.send(w, a.marshalled())
 }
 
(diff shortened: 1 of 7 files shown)


Tushar3099:
@Harkishen-Singh I have accessed the in-memory chain of a particular hash using BRMatrix and then appended the chain to the chain obtained from secondary storage and sent this to the frontend. I assume the result is showing what was expected. 
Please tell me if the approach I used is right.
Both the Testing window and Monitoring window are querying flushed data along with in-memory chains.

src/lib/api/api.go
```diff
@@ -174,8 +174,10 @@ func (a *API) Query(w http.ResponseWriter, r *http.Request) {
 		startTimestamp, endTimestamp int64
 		err                          error
 	)
-	timeSeriesPath := r.URL.Query().Get("timeSeriesPath")
-
+	hash := r.URL.Query().Get("hashKey")
+	Type := r.URL.Query().Get("type")
```
Harkishen-Singh, line 178:
```suggestion
	var (
	    hash = r.URL.Query().Get("hashKey")
	    queryType = r.URL.Query().Get("type")
	)
```

src/lib/api/api.go
```diff
@@ -174,8 +174,10 @@ func (a *API) Query(w http.ResponseWriter, r *http.Request) {
...
+	var timeSeriesPath string
+	var newBlocks []tsdb.Block
```
Harkishen-Singh, line 180: These two are used down. I suggest declaring them close to their usecase.

src/lib/api/api.go
```diff
@@ -196,25 +198,54 @@ func (a *API) Query(w http.ResponseWriter, r *http.Request) {
...
+			return
+		}
+		matrix := (*a.Matrix)[hash]
+
+		switch Type {
+		case "ping":
+			newBlocks = matrix.PingChain.Chain
+		case "jitter":
+			newBlocks = matrix.JitterChain.Chain
+		default:
```
Harkishen-Singh, line 228: Will `Type` not have `monitoring`?

src/lib/api/api.go
```diff
@@ -196,25 +198,54 @@ func (a *API) Query(w http.ResponseWriter, r *http.Request) {
...
+			timeSeriesPath = "storage/journal" + tsdb.FileExtension
+		default:
+			a.send(w, []byte("INVALID_TYPE"))
+			return
+		}
+		if ok := tsdb.VerifyChainPathExists(timeSeriesPath); !ok {
+			a.send(w, []byte("INVALID_PATH"))
+			return
+		}
+		newBlocks = []tsdb.Block{}
```
Harkishen-Singh, line 215: Why is this required?

src/lib/api/api.go
```diff
@@ -196,25 +198,54 @@ func (a *API) Query(w http.ResponseWriter, r *http.Request) {
...
-	}
 
 	// TODO: we do not capture the block streams in memory while querying yet. They are captured only when flushed.
 	// TODO: consider cmap while querying for fetching latest blocks after shifting tsdb to binary.
 
 	qry := querier.New(timeSeriesPath, "", querier.TypeRange)
 	query := qry.QueryBuilder()
 	query.SetRange(startTimestamp, endTimestamp)
-	a.Data = query.ExecWithoutEncode()
+	a.Data = query.ExecWithoutEncode(newBlocks)
```
Harkishen-Singh, line 248: I see newBlocks is always going to be empty? So is there a reason to pass this down the func?

tsdb/querier/querier.go
```diff
@@ -88,9 +88,10 @@ func (q *Query) Exec() []byte {
 
 // ExecWithoutEncode executes the Query without encoding the result to []byte,
 // rather keeps the result as default QueryResponse.
-func (q *Query) ExecWithoutEncode() QueryResponse {
+func (q *Query) ExecWithoutEncode(newBlocks []tsdb.Block) QueryResponse {
```
Harkishen-Singh, line 91: If you need blocks to be passed for querying, we should have a `FillDataBlocks()` that should do this job. The param here is not defining anything IMO.

harkishen (changes requested):

src/lib/api/api.go
Tushar3099: I will do these changes.

src/lib/api/api.go
Tushar3099: Actually, `Query` is used for both `system` information and `tests`(ping/jitter). Incase of jitter/ping I am sending the route hash to access the in-memory chains because through `BRMatrix` we can access the chains, but we cannot access the system information like that so I am just sending the flushed data for system metrics.

src/lib/api/api.go
Tushar3099: Because `Query` is used for Testing which only has Ping and Jitter for now, so I just made cases for them. I will include monitoring also for future purposes.

src/lib/api/api.go
Tushar3099: No not-neccessary as : 
```
		case "ping":
			newBlocks = matrix.PingChain.Chain
		case "jitter":
			newBlocks = matrix.JitterChain.Chain
		default:
```
`newBlocks` will have the unflushed memory-chains which are monitored and may have new blocks.

tsdb/querier/querier.go
Tushar3099: Actually, earlier in `ExecWithoutEncode`, we reloaded all the blocks from the DBpath and ran `exec` function. Now I am passing down the newBlocks(in-memory chain), which is then appended to the fetched blocks from the local storage, then we are executing `exec`. Please explain how can I implement this otherwise.

src/lib/api/api.go
Tushar3099: I am having this inference from main.go : 
```
		chainSet.Register(fmt.Sprintf("%s-ping", uHash), targetMachineCalc[uHash].Ping)
		chainSet.Register(fmt.Sprintf("%s-jitter", uHash), targetMachineCalc[uHash].Jitter)
		chainSet.Register(fmt.Sprintf("%s-fping", uHash), targetMachineCalc[uHash].FPing)
```
as targetMachineCalc[uHash] are registered in the `chainset` and also,
```
	        matrix[hash] = &utils.BRMatrix{
		        Route:        r,
		        PingChain:    targetMachineCalc[uHash].Ping,
		        JitterChain:  targetMachineCalc[uHash].Jitter,
		        FPingChain:   targetMachineCalc[uHash].FPing,
		        MonitorChain: tsdb.NewChain(pathMonitor).Init(),
		        Metrics:      endpointMetrics,
	        }
```
targetMachineCalc[uHash] is assigned to the matrix , so when : 
```
		        case "Ping":
			        workers.Ping.Iterate("stop", false)
		        case "FloodPing":
			        workers.PingF.Iteratef("stop", false)
		        case "Jitter":
			        workers.Jitter.Iterate("stop", false)
		        case "ReqResDelayMonitoring":
			        workers.Monitor.Iterate("stop", false)
		        }
```
It is used to iterate over all the routes the through `targetMachineCalc[uHash]`, in-memory chains will be updated in matrix as in worker we pass `targetMachineCalc[uHash]`,
```
	workers := &struct {
		Ping    *ping.Ping
		Jitter  *jitter.Jitter
		PingF   *ping.FloodPing
		Monitor *monitor.Monitor
	}{
		Ping:    ping.New(conf, ping.TestInterval{OfType: intervals[0].Type, Duration: *intervals[0].Duration}, &targetMachineCalc),
		Jitter:  jitter.New(conf, jitter.TestInterval{OfType: intervals[0].Type, Duration: *intervals[1].Duration}, &targetMachineCalc),
		PingF:   ping.Newf(conf, ping.TestInterval{OfType: intervals[0].Type, Duration: *intervals[0].Duration}, conf.Config.Password, &targetMachineCalc),
		Monitor: monitor.New(conf, monitor.TestInterval{OfType: intervals[2].Type, Duration: *intervals[2].Duration}, &matrix),
	}
```

Tushar3099:
@Harkishen-Singh, Can you please reply to the comments so that I can improve the PR and make the necessary changes? Thank you.

PR #428, opened 2021-03-25
formatted test fixture json

devansh42:
I've formatted test fixture JSON files, now they are making more sense at first sight.


muskankhedia:
@devansh42 Thank you for the PR. 

There is a specific motive for keeping all the data on a single line, one of them is to save the disk space that can be acquired by `\n` character and this is handled in the code, changing each file won't make a difference. So, closing this PR.

PR #430, opened 2021-03-25
add label for centos

aquibbaig:
Fixes the CI build.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -32,6 +32,6 @@ jobs:
           context: .
           file: ./Dockerfile
           platforms: |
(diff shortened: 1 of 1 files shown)


aquibbaig:
@muskankhedia The build was failing mostly due to the fact that `centos` image does not support `linux/386` architecture.
See https://hub.docker.com/_/centos.

muskankhedia:
> @muskankhedia The build was failing mostly due to the fact that `centos` image does not support `linux/386` architecture.
> See [hub.docker.com/_/centos](https://hub.docker.com/_/centos).

Ohh okay, thank you for looking into it.

muskankhedia (approved):

PR #431, opened 2021-03-25
Open browser on startup

devansh42:
Added functionality to open the dashboard in a new default browser window on startup.

### DIFF
--- a/src/main.go
+++ b/src/main.go
@@ -14,6 +14,7 @@ import (
 	"time"
 
 	"github.com/gorilla/mux"
+	"github.com/pkg/browser"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/common/log"
 	"github.com/rs/cors"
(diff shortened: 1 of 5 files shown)


src/main.go
```diff
@@ -306,6 +307,11 @@ func main() {
 		os.Exit(0)
 	}()
 	log.Infoln("Bench-routes is up and running")
+	browserableUrl := fmt.Sprintf("http://localhost%v", port)
+	err := browser.OpenURL(browserableUrl)
+	if err == nil {
+		log.Infoln("Dashbord is opened in default browser at " + browserableUrl)
+	}
 	log.Errorln(http.ListenAndServe(port, cors.Default().Handler(router)).Error())
```
ParthGala2k, line 315: Hi @devansh42, this is pretty good for a start.
But the feature that we ultimately want in this issue is to be able to start a browser of ones choice via a flag ie. like `chrome`, `firefox`, `midori`, etc. If a flag is not provided, then the default shall be used. This implementation will most likely only open the default browser of the user.

Besides we need to ensure cross compatibility of this code on all three operating systems namely Linux, macOS and Windows. All future enhancements must be done keeping in mind the inclusion of Windows since bench-routes currently does not support it as well.

Also, you haven't checked the case for when `err != nil`, which is considered good practice in golang. You must try to include that as well.

Unfortunately, before you made this pull request, I had already completed my side of the commits, it would be great if you could see that, and give a review if it works on your OS(I use ubuntu 20,04 and it works fine on it).

ParthGala2k (changes requested):
Changes mentioned in main.go are to be implemented.

PR #432, opened 2021-03-25
added new system metrics

codetheorem:
Signed-off-by: codetheorem <hrishikeshagarwalv@gmail.com>

Added new system metric to get per core cpu usgae.

### DIFF
--- a/src/metrics/system/system.go
+++ b/src/metrics/system/system.go
@@ -20,6 +20,7 @@ const (
 type CPU interface {
 	GetTotalCPUUsage() string
 	GetTotalCPUStats() gostats.TimesStat
+	GetPerCoreUsage() string
 }
 
 // Memory interface for Memory related details.
@@ -63,6 +64,19 @@ func (s *SystemMetrics) GetTotalCPUStats() gostats.TimesStat {
 	}
 
 	return stats[0]
+}
+
+func (s *SystemMetrics) GetPerCoreUsage(c chan *string) {
+	perCore, err := gostats.Percent(time.Duration(time.Second), true)
+	if err != nil {
+		panic(err)
+	}
+	str := ""
+	for i := 0; i < len(perCore); i++ {
+		currentCpu := fmt.Sprintf("%.2f", perCore[i])
+		str += currentCpu + " "
+	}
+	c <- &str
 }
 
 // MemoryStats for memory based statistics.


src/metrics/system/system.go
```diff
@@ -65,6 +66,19 @@ func (s *SystemMetrics) GetTotalCPUStats() gostats.TimesStat {
 	return stats[0]
 }
 
+func (s *SystemMetrics) GetPerCoreUsage(c chan *string) {
+	perCore, err := gostats.Percent(time.Duration(time.Second), true)
+	if err != nil {
+		panic(err)
```
Harkishen-Singh, line 72: we shouldnt panic, rather return error.

src/metrics/system/system.go
```diff
@@ -65,6 +66,19 @@ func (s *SystemMetrics) GetTotalCPUStats() gostats.TimesStat {
...
+	}
+	str := ""
```
Harkishen-Singh, line 74:
```suggestion
	var s string
```

Harkishen-Singh (changes requested):

codetheorem:
@Harkishen-Singh please review this.

src/metrics/system/system.go
```diff
@@ -1,6 +1,7 @@
 package system
 
 import (
+	"errors"
```
Harkishen-Singh, line 4: This is not required.

src/metrics/system/system.go
```diff
@@ -65,6 +67,19 @@ func (s *SystemMetrics) GetTotalCPUStats() gostats.TimesStat {
 	return stats[0]
 }
 
+func (s *SystemMetrics) GetPerCoreUsage(c chan *string) {
+	perCore, err := gostats.Percent(time.Duration(time.Second), true)
+	if err != nil {
+		errors.Unwrap(err)
```
Harkishen-Singh, line 73: I think this is a trivial statement. It returns an error that is never captured. Things like this should be sent via an error channel or use a single channel that accepts an interface and at the receiving side, to the type matching/conversion.

Harkishen-Singh (changes requested):

Harkishen-Singh:
I was expecting this to be implemented into bench-routes and not just a func addition in collectors. But anyways, we can do that in some other PR.

PR #433, opened 2021-03-26
Implement feature for web browser opening up on its own

ParthGala2k:
upon running bench routes either through `make run` or through the
executable.

This would take out the hassle from having to enter a url on
your own into the browser. Currently, the browser to be opened
has been stored as a variable, called browserName and has been
defaulted to firefox, but the ultimate goal is to be able to
pass the browser name as a flag and open a default if none is
provided.

But that might need command line parsing, one feature that is
in the pipeline after which we can come back and implement this
remaining functionality.

The changes here have been made keeping in mind that this code
must work on all the three intended operating systems namely
Linux, Windows and MacOS.

### DIFF
--- a/src/main.go
+++ b/src/main.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"os/exec"
 	"os/signal"
 	"reflect"
 	"runtime"
@@ -62,6 +63,7 @@ var (
 	pathJitter     = fmt.Sprintf("%s/jitter", storageDir)
 	pathFloodPing  = fmt.Sprintf("%s/flood-ping", storageDir)
 	pathMonitoring = fmt.Sprintf("%s/monitoring", storageDir)
+	browserName    = "firefox"
 )
 
 func main() {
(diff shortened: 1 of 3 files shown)


src/main.go
```diff
@@ -306,6 +308,29 @@ func main() {
...
 	}()
 	log.Infoln("Bench-routes is up and running")
+	url := fmt.Sprintf("http://localhost%s", port)
+	go func(url string) {
+		var err error
+		log_message := fmt.Sprintf("%s is running bench-routes at %s", browserName, url)
+
+		switch runtime.GOOS {
+		case "linux":
+			err = exec.Command(browserName, url).Start()
```
Harkishen-Singh, line 318: I am sorry, but I feel the browser types, as which browser opens is trivial for the project. This is just my opinion, but other maintainers can argue.

ParthGala2k: Certainly, it is more of a nice to have feature than need to have. But I was looking at it from the perspective of a command line only application, where for example you would be able to run bench routes in CLI mode by default, and have to pass an argument like `--web=firefox` for example in order to start it in the browser.

But then, I think opening any browser at all by default should be a natural progression from having to type it in yourself.
I would like to listen to other opinions as well, from anyone who uses the application as to how they would prefer it.

PR #435, opened 2021-03-26
Added null values for blocks

### LINKED ISSUE #242: Insert null values for blocks that dont exists in UI
currently, if a block does not exist, the graph is formed instead of discontinuing for those points. This can be done by inserting `null` in those places in the frontend - v1.1

Rits1272:
Signed-off-by: rits1272 ritikjain1272@gmail.com

File Changed: `src/main.go`

 ### Changes
 - Added `addNullBlock` function to add null values for blocks in `main.go`
- Now, let's say if bench-routes is stopped for 5 minutes and then started again, the graph won't be continuous as it used to be earlier. The graph will be now empty for the time bench-routes was stopped

[Video updated till latest commit]

https://user-images.githubusercontent.com/44135362/114286457-93217e00-9a7c-11eb-9273-f87f6dc341ff.mp4

Resolves #242

### DIFF
--- a/src/main.go
+++ b/src/main.go
@@ -64,63 +64,6 @@ var (
 	pathMonitoring = fmt.Sprintf("%s/monitoring", storageDir)
 )
 
-// EXPERIMENTAL (ref: https://github.com/bench-routes/bench-routes/issues/242)
-
-func GetTimeStampCalc(sub int) string {
-	t := time.Now().Add(time.Second * time.Duration(-sub)) // getting the timestamp `sub` seconds previous
-
-	return s(t.Year()) + "|" + s(t.Month()) + "|" + s(t.Day()) + "|" + s(t.Hour()) + "|" +
-		s(t.Minute()) + "|" + s(t.Second())
-}
-
-func s(st interface{}) string {
-	return fmt.Sprintf("%v", st)
-}
-
-func addNullBlock(chain *tsdb.Chain, path, nullEncodedData, metricType string) {
-	chainReadOnly := tsdb.ReadOnly(path).Refresh() // getting the chain from the disk
-	bstream := chainReadOnly.BlockStream()
-
-	blockStream := *bstream
-
-	lengthBlockStream := len(blockStream)
-
-	if lengthBlockStream == 0 {
-		return
-	}
-
-	block := blockStream[lengthBlockStream-1] // getting the most recent block added to the disk
-
-	blockNormTime := block.GetNormalizedTime()
-	curNormTime := tsdb.GetNormalizedTimeCalc()
-
-	var blockChain []tsdb.Block // for null blocks
-	sub := 0
-	time := int64(processCollectionScrapeTime) // time is in nano seconds
-
-	/*
-		Starting from the current normalized time, we go back
-		and add null blocks until we get past the block
-		normalized time which is the most recent block added
-		at the time when bench-routes was stopped
-	*/
-	for curNormTime > blockNormTime {
-		b := tsdb.GetNewBlock(metricType, nullEncodedData)
-		b.NormalizedTime = curNormTime
-		b.Timestamp = GetTimeStampCalc(sub)
-		blockChain = append(blockChain, *b)
-		sub = sub + int(time/1000000000)
-		curNormTime -= time
-	}
-
-	// append blocks to the chain in reverse order
-	blockChainLength := len(blockChain)
-
-	for a := blockChainLength - 1; a >= 0; a-- {
-		chain.Append(blockChain[a])
-	}
-}
-
 func main() {
 	if len(os.Args) > 2 && os.Args[2] != "" {
 		enableProcessCollection, _ = strconv.ParseBool(os.Args[2])
@@ -235,9 +178,6 @@ func main() {
 		log.Infoln("initialized system-metrics...", time.Since(p))
 		chainSet.Register(chain.Name, chain)
 
-		nullEncodedData := "0|0|0|0|0|0|0|0|0|0|0|0|0|0"
-		addNullBlock(chain, systemMetricsPath, nullEncodedData, "sys")
-
 		for {
 			// collections for cpu, memory and disk run independently and are
 			// time dependent. Hence, running these serially will take more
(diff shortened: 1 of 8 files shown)


Rits1272:
Kindly review @Harkishen-Singh

src/main.go
```diff
@@ -64,6 +64,63 @@ var (
 	pathMonitoring = fmt.Sprintf("%s/monitoring", storageDir)
 )
 
+// EXPERIMENTAL (ref: https://github.com/bench-routes/bench-routes/issues/242)
```
Harkishen-Singh, line 67: The experimental code along with GetTimeStampCalc(), s, addNullBlock() should not be in main.go. This should go to tsdb dir and when chains start (or if any chain is loaded into memory the first time, this process will happen), this operation should be done there.

src/main.go
```diff
@@ -64,6 +64,63 @@ var (
...
+		at the time when bench-routes was stopped
+	*/
+	for curNormTime > blockNormTime {
+		b := tsdb.GetNewBlock(metricType, nullEncodedData)
+		b.NormalizedTime = curNormTime
+		b.Timestamp = GetTimeStampCalc(sub)
+		blockChain = append(blockChain, *b)
+		sub = sub + int(time/1000000000)
+		curNormTime -= time
+	}
```
Harkishen-Singh, line 114: I do not understand why do we need to add so many 0 blocks?

Some points to note:
1. A zero block is not a null block. Doing `0|0|0|0|0|0|0|0|0|0|0|0|0|0` is not the solution. I like the idea, but that will be far too expensive on memory and in HTTP requests. Imagine a chain which was written last month, and so the blockNormTime will be 1 month behind, then you can imagine the number of blocks that will be added. This will not be scalable and be actually too expensive.
2. For making graphs null, you should first see how will the graphs respect this efficiently, Fortunately, apexcharts have null values support. See this https://apexcharts.com/javascript-chart-demos/area-charts/chart-with-missing-data-or-null-values/. So if I am not wrong, just inserting a single `null` in value and the timestamp and type being the same (meaning non-null), in the very start of a chain should solve this. The querier can then before passing to UI, make some tweaks plus there needs to be a bit of changes in our parsing way in graphs which can be a good amount of work, but still doable. I guess in UI, the place where we segregate a single response into multiple data values for different graphs can be tweaked in a way that, as it sees a `"null"`, it inserts  `"null"` into those multiple data values and then this should be good to go.

src/main.go
```diff
@@ -178,6 +235,9 @@ func main() {
 		log.Infoln("initialized system-metrics...", time.Since(p))
 		chainSet.Register(chain.Name, chain)
 
+		nullEncodedData := "0|0|0|0|0|0|0|0|0|0|0|0|0|0"
```
Harkishen-Singh, line 238: This is a zero block. Null block means a block with value as `null`. Please make sure you go through the above mentioned null example in apexcharts before you make changes here.

src/main.go
```diff
@@ -221,6 +281,9 @@ func main() {
 			log.Infoln("initialized journal-metrics...", time.Since(p))
 			chainSet.Register(chain.Name, chain)
 
+			nullEncodedData := "0|0|0|0|0|0"
+			addNullBlock(chain, journalMetricsPath, nullEncodedData, "journal")
```
Harkishen-Singh, line 285: Having addNullBLock in tsdb will make life simpler by not caring about type, since as the chain is formed, it should handle the types and other related information for null values. That way, its much easier to use, than always keeping in mind to call an addNullBlock() that does this, and passing all the necessary args.

Harkishen-Singh (changes requested):
Thank you for taking this up. This is indeed something not simple. The idea is good, but there are some concerns related to scalability. Let me know if you face any issues with the suggestions.

src/main.go
Rits1272: @Harkishen-Singh I just want to make sure that I have understood things correctly before making the next commit. 

1. As you said, the `addNullBlock` method along with some other methods will be added in the `tsdb` instead of the `main.go` file at the time when the chain starts.

2. The `zero` blocks that have been added will be changed to `null` blocks since the `apex charts` provides us a way to handle the `null` value as you mentioned. 

3. This zero block `0|0|0|0|0|0|0|0|0|0|0|0|0|0` is to be replaced with a single `null` and a `timestamp`, and thereby making the necessary changes in the `parser` and `UI` to handle this change.

Please correct me I am wrong or missed out on something. Thank you :)

Harkishen-Singh: That looks correct

Rits1272:
Kindly review @Harkishen-Singh. I have made the changes as per the discussion. (Redundant console.log will be removed in the next changes if any)

tsdb/db.go
```diff
@@ -194,6 +195,55 @@ func VerifyChainPathExists(chainPath string) bool {
...
+		at the time when bench-routes was stopped
+	*/
+	for curNormTime > blockNormTime {
+		b := GetNewBlock(blockType, "null")
+		b.Timestamp = CalcTimeStamp(sub)
+		b.NormalizedTime = curNormTime
+		blockChain = append(blockChain, *b)
+		sub = sub + int(time/1000000000)
+		curNormTime -= time
+	}
```
Harkishen-Singh, line 238: Aren't we still adding so many null blocks? The idea was to just add a single null block only and not so many null blocks. the UI will consider the null block and apexcharts will automatically remove the graph until the next value (which will be immediately after this null block).

imagine time of say -> 1, 2, 3, 4, 5, 11, 12, 13, 14

then, we do

1, 2, 3, 4, 5, null, 11, 12, 13, 14

Rits1272: Thanks for the review @Harkishen-Singh. I have made the changes. Kindly check

Harkishen-Singh (approved):
I see null blocks support is just on the dashboard and not on the monitoring graphs. I will merge this for now, but kindly make sure that when you implement a feature/fix, it should be implemented across the entire project.

Rits1272:
Sure @Harkishen-Singh. First of all, thank you for the reviews and the guidance provided. I will surely implement a feature/fix for the same soon.

Issue #61, opened 2019-10-07
Add individual go-routine sleep for ping


Harkishen-Singh:
The `sleep` in main go-routine handling ping makes the entire thread sleep only when all the ping operation on the routes mentioned in the config file has been done _(a dependent behaviour for each ping operation)_. This is kind of a wrong logic as we need to pause the thread for the specified interval specifically for the particular IP address.

The correct approach would be to bring this inside the individual go-routine would be logically correct as only those threads would be adjourned whose ping operation has been completed, thus being independent of other ping subprocess running in the background.

Tushar3099:
@Harkishen-Singh I would like to work on this issue.

Harkishen-Singh:
Please go ahead

devansh42:
Hey @Harkishen-Singh  I'd like to work on this issue, @Tushar3099 and I can collaborate on this.

Harkishen-Singh:
You are most welcome.

devansh42:
@Tushar3099 What's your views on [#441 ](https://github.com/bench-routes/bench-routes/pull/441) issues.

Tushar3099:
@devansh42 I went through your PR a little bit, I will dig in later as right now I am working on other issues. It seems there is no other task left to do in the PR, so I will leave the collaboration up to you. I will love to review the PR in detail in some time.
From now onwards, Kindly avoid raising PR on issues that are already assigned to others without notifying as this may lead to a waste of effort.

Tushar3099:
@devansh42, your approach may be right but using outside for loop to continuously check whether the channel is closed or not doesn't seemed to be a very promising approach to me. Can we just run goroutines for each of the routes, which will in itself run infinitely? When these go-routines return, pinging will stop. I am not sure, but this may be an approach.
I have raised a PR following my approach.
Also, please avoid raising PRs which are already assigned to others without prior notification.

PR #441, opened 2021-03-28
Ping sleep enhancement

devansh42:
Added functionality of sleep in individual ping go-routine, But this gives rise to the new problem as the outermost for loop will now run super-fast and burn unnecessary CPU cycles, the function *utils.VerifyConnection* also would make a network call in each iteration.
So @Harkishen-Singh, Is this the desirable behavior.

### DIFF
--- a/src/lib/modules/ping/ping.go
+++ b/src/lib/modules/ping/ping.go
@@ -77,7 +77,30 @@ func (ps *Ping) setConfigurations() {
 	ps.perform(pingInterval)
 }
 
+// pingerMap holds the closing channel for current set of pingable machines workers
+type pingerMap map[string]chan struct{}
+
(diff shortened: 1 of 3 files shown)


Tushar3099:
@devansh I have been working on the issue for a while #61 but due to merging conflicts of other PR, I had some delay. If you were interested in working on this issue you could have said on the issue itself. Raising PR like this without notifying results in wasted engineering efforts. Please do care for this.

devansh42:
Sure @Tushar3099 I'd mind your suggestions. What about working together on this issue :)

Tushar3099:
@devansh42  Thanks for inviting me to collaborate on this issue, I am studying this issue right now. We can do the rest of discussion on the issue #61.

PR #442, opened 2021-03-29
Individual go-routine sleep

### LINKED ISSUE #61: Add individual go-routine sleep for ping

Tushar3099:
Signed-off-by: Tushar3099 <tusharneogi30@gmail.com>
fixes :  #61 

- [x]  Enhancement

**Changes**
-----
- In `ping.go` I have modified the perform function.
- Earlier perform function was running infinitely for all set of routes with sleeping duration.
- Now the perform functions execute `pingExec` function for all sets of routes and waits for them to complete.
- `pingExec` function runs infinitely for every route independent of other routes running in the background with sleep duration.
- `pingExec` executes the `ping` function as earlier.

**Expected**
-----
- Now ping for each route is executed independently from other routes running in the background.

Additional Context
-----
- The PR is not fully refactored. Once the approach is confirmed, will do the necessary changes.

### DIFF
--- a/src/lib/modules/ping/ping.go
+++ b/src/lib/modules/ping/ping.go
@@ -88,6 +88,7 @@ func (ps *Ping) perform(pingInterval TestInterval) {
 }
 
 func (ps *Ping) pingExec(urlRaw, machineHash string, packets int, wg *sync.WaitGroup, pingInterval TestInterval) {
+	defer wg.Done()
 	for {
 		switch ps.localConfig.Config.UtilsConf.ServicesSignal.Ping {
 		case "active":
@@ -102,11 +103,9 @@ func (ps *Ping) pingExec(urlRaw, machineHash string, packets int, wg *sync.WaitG
 			}
 		case "passive":
 			// terminate the goroutine
-			wg.Done()
 			return
 		default:
 			log.Warnln("invalid service-state value of ping")
-			wg.Done()
 			return
 		}
 
--- a/src/lib/modules/ping/ping.go
+++ b/src/lib/modules/ping/ping.go
@@ -78,6 +78,16 @@ func (ps *Ping) setConfigurations() {
 }
 
 func (ps *Ping) perform(pingInterval TestInterval) {
+	defer log.Infoln("terminating ping goroutine")
+	var wg sync.WaitGroup
+	wg.Add(len(*ps.targets))
+	for machineHash, machineIP := range *ps.targets {
+		go ps.pingExec(machineIP.IPDomain, machineHash, 3, &wg, pingInterval)
+	}
+	wg.Wait()
+}
+
+func (ps *Ping) pingExec(urlRaw, machineHash string, packets int, wg *sync.WaitGroup, pingInterval TestInterval) {
 	for {
 		switch ps.localConfig.Config.UtilsConf.ServicesSignal.Ping {
 		case "active":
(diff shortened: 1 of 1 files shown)


Tushar3099:
@Harkishen-Singh please confirm if the approach is right. Will do the necessary changes.

src/lib/modules/ping/ping.go
```diff
@@ -78,26 +78,35 @@ func (ps *Ping) setConfigurations() {
...
+				waitgroup.Wait()
 			}
 		case "passive":
 			// terminate the goroutine
-			log.Infoln("terminating ping goroutine")
+			wg.Done()
 			return
 		default:
 			log.Warnln("invalid service-state value of ping")
+			wg.Done()
```
chetanyakan, line 109: Ideally, we should use `defer wg.Done()` as the first line of this function instead of calling `wg.Done()` at each return statement.

Right now, as `wg.Done()` is not called before line 123, there is a possibility that the parent function will never stop executing.

Tushar3099: Thanks for the review, @chetanyakan, but actually that is what the expected behavior should be. Also, that is the reason I didn't put wg.Done() in defer statement.
In the case of active, the parent function is not supposed to stop, instead, sleep for the specified duration and execute again till the status is not turned `passive`.

chetanyakan: @Tushar3099 Don't you think there should be an equal number of `wg.Add()` and `wg.Done()` calls or otherwise the program won't stop even if there's nothing to execute?

P.S. In case of active, it will keep working in the same manner as of right now, even when you call `defer wg.Done()` before the for loop.

Tushar3099: @chetanyakan Please be little specific here : 

> @Tushar3099 Don't you think there should be an equal number of wg.Add() and wg.Done() calls or otherwise the program won't stop even if there's nothing to execute?

I don't get what you want to say : 

> P.S. In case of active, it will keep working in the same manner as of right now, even when you call defer wg.Done() before the for loop.

- We are adding wait groups equal to the number of targets and then running the pingExec functions.
- We want them to run infinitely with sleep durations until the status is passive.
- In the case of active we do not want the function to stop, also it is obvious if we put defer statement outside the loop it will never be executed in case of active as it will never come out of the loop.
- We can do one thing to run a defer statement outside the loop and the return from the loop in case of default or passive, but that would also work the same as of right now.

If there is a possible way where we would want the function to stop but it won't stop please point it out. But in the case of active, we don't want the function to stop.

chetanyakan: > We can do one thing to run a defer statement outside the loop and the return from the loop in case of default or passive, but that would also work the same as of right now.

I think it would not work the same as of right now. But please correct me if I am wrong.
Please look at the code below (lines 113 - 124)
```go
		intrv := time.Duration(*ps.localConfig.Config.Interval[0].Duration)
		switch pingInterval.OfType {
		case "hr":
			time.Sleep(intrv * time.Hour)
		case "min":
			time.Sleep(intrv * time.Minute)
		case "sec":
			time.Sleep(intrv * time.Second)
		default:
			log.Infoln("invalid interval-type for ping")
			return
		}
```
What would happen if the program returns in the `default` case of `pingInterval.OfType`?

1. Would `wg.Done()` be called or not?
2. Should `wg.Done()` be called here to make sure `wg.Wait()` on line 87 can finish executing?

Tushar3099: That's absolutely right. Thanks for pointing out, we would require one more wg.Done(), before this default statement. However, I  think I should write wg.Done() outside loop with defer so that we don't have to put wg.Done() everywhere.

Tushar3099:
@aquibbaig @Harkishen-Singh please tell me if the applied approach is correct.

Tushar3099:
Please review my PR @Harkishen-Singh @aquibbaig  @muskankhedia

Issue #418, opened 2021-03-22
Arrow Pointer inside "Tests"  menu option in sidebar should be directed appropriately.


tasneemkoushar:
**Describe the bug**
Before we click on the "Tests" menu option in the sidebar, the arrow points to the upward direction. After we click on "Tests", the options like "Ping", "FloodPing", "Jitter" appear in downward submenu format, so the arrow pointer should initially be pointing downward and after expanding the "Tests" menu, it must be pointing upward.

**To Reproduce**
1. Expand The sidebar
2. Observe the arrow pointer of the "Tests" menu
3. Click on the "Tests" menu 
4. Now observe the arrow Pointer direction
5. Now we can see the arrow hasn't changed its direction in response to the submenu options.

**Expected behavior**
There is no technical fault but the arrow direction change from downward to upward and vice-versa will improve the aesthetic view of the "Tests" menu and enhance the user interface.

**Screenshots**
![1-view](https://user-images.githubusercontent.com/58592306/112040820-aae79f80-8b6b-11eb-934d-1700d28dedf2.jpg)
![2-view](https://user-images.githubusercontent.com/58592306/112040844-b1761700-8b6b-11eb-9a4e-95c8483dc93b.jpg)

**Desktop (please complete the following information):**
 - OS: Windows 10

**Additional context**
I want to work on this, please assign this issue to me.

tasneemkoushar:
Please assign this issue to me.

aquibbaig:
@tasneemkoushar I don't think it matters as much but go ahead!!

tasneemkoushar:
> @tasneemkoushar I don't think it matters as much (since we're going to move to v2.0) but sure, go ahead!!

Alright! Thanks, I will start working.

tasneemkoushar:
https://github.com/bench-routes/bench-routes/pull/449
My PR passed both the Automation checks, Please review it.

PR #447, opened 2021-03-29
Test expand icon fix

tasneemkoushar:
**#418** 
**bug** - Boolean "open" in "dashboard/v1.1/src/layouts/Sidebar.tsx" in line 126  never changed its value, so the transition between
ExpandMoreIcon and ExpandLessIcon was not functional.

**Fix** - I added a new state hook "expand" in line 84, set it's value initially to "true" and replaced this state with "open" in line 126, then with the function "showTestList"  in line 86, the state  of "expand" is changing to its opposite value "!expand"  in state function "setExpand" declared on line 84.

### DIFF
--- a/dashboard/v1.1/src/layouts/Sidebar.tsx
+++ b/dashboard/v1.1/src/layouts/Sidebar.tsx
@@ -85,7 +85,7 @@ const Sidebar: FC<SidebarProps> = ({ handleDrawerClose, open }) => {
 
   const showTestList = () => {
     setTestListOpen(!testListOpen);
-    setExpand(false);
+    setExpand(!expand);
   };
 
   const handleListItemClick = (event, index) => {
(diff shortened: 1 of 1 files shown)


dashboard/v1.1/src/layouts/Sidebar.tsx
```diff
@@ -81,8 +81,11 @@ const Sidebar: FC<SidebarProps> = ({ handleDrawerClose, open }) => {
   // Sidebar element
   const [testListOpen, setTestListOpen] = useState(false);
   const [selectedIndex, setSelectedIndex] = React.useState(0);
+  const [expand, setExpand] = useState(true);
+
   const showTestList = () => {
     setTestListOpen(!testListOpen);
+    setExpand(!expand);
```
chetanyakan, line 88:
```suggestion
    setTestListOpen(isOpen => !isOpen);
    setExpand(prevExpand => !prevExpand);
```

PR #448, opened 2021-03-29
Hide quick input Icon on Quick input page

### LINKED ISSUE #419: Quick Route Input should not be on there Quick Route input page.

ruddi10:
Signed-off-by: ruddi <rudrakshaggarwalsachin@gmail.com>

fixes #419 
Screenshot from 2021-03-30 01-46-55

### DIFF
--- a/dashboard/v1.1/src/layouts/QuickInputFab.tsx
+++ b/dashboard/v1.1/src/layouts/QuickInputFab.tsx
@@ -1,5 +1,6 @@
 import { Fab, makeStyles, Tooltip } from '@material-ui/core';
 import { PostAdd as PostAddIcon } from '@material-ui/icons';
+import { withRouter } from 'react-router-dom';
 import React from 'react';
 import { Link } from 'react-router-dom';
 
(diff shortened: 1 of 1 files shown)


ruddi10:
@muskankhedia  @aquibbaig kindly review

dashboard/v1.1/src/layouts/QuickInputFab.tsx
```diff
@@ -22,8 +23,12 @@ const useStyles = makeStyles(theme => ({
 }));
 
 // Floating Action Button for Quick Input
-const QuickInputFab = () => {
+const QuickInputFab = props => {
   const classes = useStyles();
+  const { location } = props;
```
chetanyakan, line 28: Can we use `useLocation` instead?

https://reactrouter.com/web/api/Hooks/uselocation

ruddi10: @chetanyakan any particular problem with this approach is so we can go with your approach

aquibbaig: Let's keep this as it is, for now, we can cautiously work on stuff in `v2.0`

dashboard/v1.1/src/layouts/QuickInputFab.tsx
```diff
@@ -40,4 +45,4 @@ const QuickInputFab = () => {
   );
 };
 
-export default QuickInputFab;
+export default withRouter(QuickInputFab);
```
aquibbaig, line 48: Why do we need to wrap this component with `withRouter`? If it is to access `location` props inside this component, let's add this logic to some top level component which can pass these values to child components

aquibbaig (changes requested):

Issue #451, opened 2021-03-30
In "/quick-input" endpoint passing http method apart from GET POST freezes goroutine


Sanketwable:
**Describe the bug**
Their is no `ResponseWrapper` object send to `response` channel.

**To Reproduce**
Steps to reproduce the behaviour:
https://github.com/bench-routes/bench-routes/blob/5855aadcf621352b2fea6ae49a441af8c74befe4/src/lib/api/api.go#L291-L314

**Expected behaviour**
There should be `ResponseWrapper` Object send to channel in order to terminate the HandlerFunc.

**Approach**
I will start another ***goroutine*** for default case in switch and send `ResponseWrapper` object to `response` channel

aquibbaig:
@Sanketwable Looks good to me, lets wait for another set of reviews.

PR #452, opened 2021-03-30
Added new goroutine to send ResponseWrapper to response channel

Sanketwable:
fixes https://github.com/bench-routes/bench-routes/issues/451

### DIFF
--- a/src/lib/api/api.go
+++ b/src/lib/api/api.go
@@ -307,7 +307,10 @@ func (a *API) QuickTestInput(w http.ResponseWriter, r *http.Request) {
 	case "POST":
 		go req.Send(request.POST, response)
 	default:
-		fmt.Printf("invalid request method: %s\n", t.Method)
+		go func() {
+			fmt.Printf("invalid request method: %s\n", t.Method)
+			response <- request.ResponseWrapper{ReponseStringified: "Invalid method", Status: 404}
+		}()
 	}
 	a.Data = <-response
 	a.send(w, a.marshalled())


src/lib/api/api.go
```diff
@@ -307,7 +307,10 @@ func (a *API) QuickTestInput(w http.ResponseWriter, r *http.Request) {
 	case "POST":
 		go req.Send(request.POST, response)
 	default:
-		fmt.Printf("invalid request method: %s\n", t.Method)
+		go func() {
+			fmt.Printf("invalid request method: %s\n", t.Method)
+			response <- request.ResponseWrapper{ReponseStringified: "Invalid method", Status: 404}
```
aquibbaig, line 312: - "When is the default method being called?" Might as well comment that above this.
- Should we call the `ResponseWrapper` on POST requests as well? What makes the default case so special?

aquibbaig (changes requested):

src/lib/api/api.go
Sanketwable: In case of POST and GET request. We are calling req.Send() in this function we are already passing Response Wrapper Object.
Default method will be called when the payload string with key "method " will have any other value apart from "GET" and "POST"

Thanks

Sanketwable: @aquibbaig can I add comment this way:
default: //Default will be called when the payload string with key "method " will have any othe...
go func() {
			fmt.Printf("invalid request method: %s\n", t.Method)
			response <- request.ResponseWrapper{ReponseStringified: "Invalid method", Status: 404
}

aquibbaig: @Sanketwable Looks good to me, lets wait for another set of reviews.

Issue #460, opened 2021-04-05
We Should add the Pagination for "/get-config-routes" endpoint.


Sanketwable:
In the `/get-config-routes` endpoint we are passing all the routes registered in ***local-config.yml***.
For now we are dealing with 10-20 request so it takes very less time for the response to get the JSON payload. But with increase in routes. It will be take huge time to get the response. 
So what i think is can we have **Pagination** in `GetConfigRoutes` function so that we will have only limited no of routes to be passed as a payload as a response. 
So the required changes will be:
1. Having routeID for every route with auto-increment starting from one. 
2. Changing endpoint format to pass startID and endID as a param to request.

Sanketwable:
@Harkishen-Singh @aquibbaig @muskankhedia I want to work in this issue.

Sanketwable:
### Approach:
1. Change in local-config.yml file to have ID for every request with auto-increment starting with 1 and one extra variable to have count of all the requests.
```
-count:23
-id:1 
 method: GET
 url: https://www.google.co.in/search
 headers:
 - type: Content-Type
   value: application/json
 params:
 - name: q
   value: bench-routes
 body: []
 labels:
 - google
 ```

2. In the GetConfigRoutes Func  accepting the URL parameters. With keys startindex and endindex. ``` r.URL.Query().Get("start_index") and r.URL.Query().Get("end_index") ```
3. Change in AddRouteToMonitoring Func to add auto increment index to all the newly added routes.

PR #463, opened 2021-04-05
Adding pagination feature

Sanketwable:
Fixes https://github.com/bench-routes/bench-routes/issues/460


Sanketwable:
Now the endpoint to fetch the Routes changes to `http://localhost:9990/get-config-routes?start_index=0&next=100`

Sanketwable:
@Harkishen-Singh @aquibbaig @muskankhedia Please review

aquibbaig:
Looks like a really good feature, will do a full-fledged review soon

Sanketwable:
> Looks like a really good feature, will do a full-fledged review soon, please check why the build fails

Actually on Mac it seems fine. I will check it once on linux. Thanks

Harkishen-Singh:
I think we can simplify `start_index` => `start` and `next` => `count` or use an end (with implementation change if necessary. PS: I haven't gone through the code)

Sanketwable:
> I think we can simplify `start_index` => `start` and `next` => `count` or use an end (with implementation change if necessary. PS: I haven't gone through the code)

OK I will change the param names. 
In case we use end we will not be sure that the total number of routes passed in the payload will be exactly what is needed. This error will occur because when we delete the route we are not refreshing the config data. And one of the RouteID will be missing.
For eg: RouteID 1 to RouteID 20 exists, and we delete Route with RouteID 13. In this case the RouteID 13 will be missing. And in case we request for routes from routeID 10 to routeID 15. Then we will receive only 4 routes rather than 5.

Issue #464, opened 2021-04-06
Maximum Response size reached while `delete-route` endpoint hit


Sanketwable:
**Describe the bug**
When the size of Routes increases in the Local-config.yml. And after this if we delete the route. We are not able to get the response as Response size is reached.

**To Reproduce**
Steps to reproduce the behaviour:
1. Populate the local-config.yml with more routes
2. Hit the delete-route endpoint
3. See error

**Expected behaviour**
As for now Delete Route returns all the Route as an Json Payload. But It should only return message : "The route is deleted" or an error : ("Route not deleted due to " + error message)

**Screenshots**
<img width="1440" alt="Screenshot 2021-04-06 at 9 00 27 AM" src="https://user-images.githubusercontent.com/43716242/113661877-93102f80-96c4-11eb-8a8e-617455ece75d.png">

Sanketwable:
@aquibbaig @muskankhedia @Harkishen-Singh I want to work on this

### Approach:

1. I will change the response Payload :
2. a.Data = "deleted successfully " in case of success and a.Data = "error message " in case of any error occurs.
3. I will change the status code to 200 (status OK ) if deleted and 404 if error occurs.

PR #465, opened 2021-04-06
Fixed Delete route issue

Sanketwable:
Fixes https://github.com/bench-routes/bench-routes/issues/464

### DIFF
--- a/src/lib/api/api.go
+++ b/src/lib/api/api.go
@@ -523,19 +523,34 @@ func (a *API) DeleteConfigRoutes(w http.ResponseWriter, r *http.Request) {
 	if err := decoder.Decode(&req); err != nil {
 		panic(err)
 	}
+	var updated bool
+	updated = false
 	for i, route := range a.config.Config.Routes {
 		if route.URL == req.ActualRoute {
 			a.mux.Lock()
 			a.config.Config.Routes = append(a.config.Config.Routes[:i], a.config.Config.Routes[i+1:]...)
+			updated = true
 			a.mux.Unlock()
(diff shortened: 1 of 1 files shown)


Sanketwable:
@Harkishen-Singh @aquibbaig @muskankhedia Please review

src/lib/api/api.go
```diff
@@ -523,19 +523,34 @@ func (a *API) DeleteConfigRoutes(w http.ResponseWriter, r *http.Request) {
 	if err := decoder.Decode(&req); err != nil {
 		panic(err)
 	}
+	var updated bool
+	updated = false
```
aquibbaig, line 527: `found := false`

aquibbaig (commented):
A couple of questions:
- In what case is "the route not present but still shows up in the Config screen"? What happens when the user presses delete route and it says "Route not found"? I sense an inconsistency here. Say there is data corruption in the config file, the question is how do we handle that? Do we force a refresh of the config screen so that we trigger a `/get-config`?

src/lib/api/api.go
Sanketwable: This is a Backend correction. In case of BenchCLI it will be very helpful.
As in case user is using CLI he will pass the Route or RouteID to delete the route from Config file and incase if it isn't present in server(spelling mistake while making request to server) will return the error message that the route you entered is not present and user will identify his mistake and again make a proper delete request the server. If in case this error is not handled user will get message that route has been deleted even though the route wasn't deleted.
In case of UI there is no issue.

Issue #466, opened 2021-04-06
AddRoute Takes lot time to add new Route to local-config.yml


Sanketwable:
### Issue
1. In the Write Function in config.go it's rewriting everything to file again despite of having this data in the file.
2. The Write Function should append data to file.
3. Due to Rewriting everything to file it takes a lot of time.

Screenshot:
<img width="1440" alt="AddNewRoute" src="https://user-images.githubusercontent.com/43716242/113675835-63b7ed80-96d9-11eb-8ab2-21c6e9bf6dd3.png">

I have tested it on 1012000 routes stored in local-config. And it takes 2 min 10 sec to add one route to local-config.
130 sec Is a lot of time to save the new route.

Sanketwable:
@Harkishen-Singh @aquibbaig @muskankhedia I want to work on this issue.

### Here's my Approach:
1. I will rewrite a new function for Adding route to local-config.yml.
2. This function will append the Route to the existing file and not rewrite whole local-config.yml
3. The function `*os.File.Write` will be used to append the Route to file instead of `ioutil.WriteFile`

PR #467, opened 2021-04-06
Add Route optimisation

Sanketwable:
fixes https://github.com/bench-routes/bench-routes/issues/466


Sanketwable:
@Harkishen-Singh @muskankhedia @aquibbaig Please review

src/lib/config/config.go
```diff
@@ -122,6 +123,26 @@ func (inst *Config) Write() (bool, error) {
 
 	return true, nil
 }
+func (inst *Config) AppendWrite(route []Route) (bool, error) {
```
aquibbaig, line 126: Interesting! Does this function return a boolean that is never used?

aquibbaig (approved):
Let's drop the boolean from the function and we're good to go!

Sanketwable (commented):
Removed the boolean return

PR #449, opened 2021-03-30
Test expand icon fix

tasneemkoushar:
#418
bug - Boolean "open" in "dashboard/v1.1/src/layouts/Sidebar.tsx" in line 126 never changed its value, so the transition between
ExpandMoreIcon and ExpandLessIcon was not functional.

Fix - I added a new state hook "expand" in line 84, set it's value initially to "true" and replaced this state with "open" in line 126, then with the function "showTestList" in line 86, the state of "expand" is changing to "!prevExpand" in state function "setExpand" declared on line 84. Besides, I also made changes in "setTestListOpen" 87 in line  function.

PR #468, opened 2021-04-06
Fixed issue #418

tasneemkoushar:
#449 
Fixed Toggling of ExpandMore and ExpandLess icon in "tests" sidebar menu.

bug - Boolean "open" in "dashboard/v1.1/src/layouts/Sidebar.tsx" in line 126 never changed its value, so the transition between
ExpandMoreIcon and ExpandLessIcon was not functional.

Fix - I added a new state hook "expand" in line 84, set it's value initially to "true" and replaced this state with "open" in line 126, then with the function "showTestList" in line 86, the state of "expand" is changing to "!prevExpand" in state function "setExpand" declared on line 84. Besides, I also made changes in "setTestListOpen" 87 in line function.


aquibbaig:
Thanks for this PR. Please note that we do not want to have multiple commits over the same change. So, what can we do to avoid it is as follows:
```
1. git rebase  -i HEAD~4 (the number of commits from your HEAD)
2. On the prefix, just type 'pick' or letter 'p' for first commit and 'squash' or letter 's' for others.
3. This will convert all your commits into a single commit. You can also have a single message for all those commits.
```

Else you can follow any corresponding tutorial on Google (Eg: https://www.git-tower.com/learn/git/faq/git-squash/)

aquibbaig:
@tasneemkoushar Also the title of this PR says "Fixed issue #418" and in the description, you have referenced #449, so that is confusing. In open-source, it has the potential to cause disaster later on. How?!

Imagine a case where I am facing an issue in my open-source project and I google it. It returns the issue you referenced as a search result. Now when I click on it and view the contents and if it is referenced incorrectly, it will cause a lot of time wasted, isn't it?

Issue #88, opened 2019-10-15
design a tool to check the the validity of configuration file


Harkishen-Singh:
**Is your feature request related to a problem? Please describe.**
Currently, the user has no idea if the configuration file made by him is valid or not, depending on the standards and the fields which are required for a valid loading of the parser.

**Describe the solution you'd like**
An independent tool can be developed which has the sole purpose of checking the config file. If on erros, it can suggest the errors and the overcoming measures.

**Additional context**
The business logic _(only)_ of this tool should be integrated with the `init()` function so that the config file could be scanned right before start of the application.

kelpikz:
hey @Harkishen-Singh, I would like to work on this issue

muskankhedia:
@kelpikz Please go ahead.

Tushar3099:
@muskankhedia, I would like to work on this issue if @kelpikz is not working on it.

muskankhedia:
@kelpikz Can you please update us with the work you have done till now on this issue?

kelpikz:
Sorry for the late reply, u can go ahead @Tushar3099

Sanketwable:
@Tushar3099 If you are not working on this can I work on this issue?

Tushar3099:
Yes @Sanketwable you can proceed with this issue, as I am working on other issues.

PR #470, opened 2021-04-07
Validity of Configuration file

Sanketwable:
Fixes https://github.com/bench-routes/bench-routes/issues/88
@Harkishen-Singh @aquibbaig @muskankhedia please review

### DIFF
--- a/src/lib/config/config-validator.go
+++ b/src/lib/config/config-validator.go
@@ -1,61 +1,130 @@
 package parser
 
 import (
+	"errors"
+	"fmt"
 	"strconv"
 
+	valid "github.com/asaskevich/govalidator"
 	"github.com/prometheus/common/log"
 )
 
 // ValidateIntervalProp validates the `test_interval` property
 // in the configuration file.
-func ValidateIntervalProp(intervals []Interval) {
+func ValidateIntervalProp(intervals []Interval) []error {
+	services := map[string]bool{
+		"ping":       true,
+		"jitter":     true,
+		"monitoring": true,
+	}
+	durationType := map[string]bool{
+		"sec": true,
+		"min": true,
+		"hr":  true,
+	}
+	intervalErr := []error{}
 	if len(intervals) == 0 {
-		log.Errorln("`test_interval` property is missing.")
+		intervalErr = append(intervalErr, errors.New("`test_interval` property is missing"))
 	} else {
+		if !services["ping"] {
+			intervalErr = append(intervalErr, fmt.Errorf("ping service missing in the rest_interval"))
+		}
+		if !services["jitter"] {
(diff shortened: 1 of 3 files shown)


src/lib/config/config-validator.go
```diff
@@ -1,61 +1,131 @@
...
 func (inst *Config) Validate() bool {
+	er := []error{}
 	config := *inst.Config
-	ValidatePasswordProp(config.Password)
-	ValidateIntervalProp(config.Interval)
-	ValidateUtilsConf(&config.UtilsConf)
+	err := ValidatePasswordProp(config.Password)
+	if err != nil {
+		er = append(er, err)
+	}
```
aquibbaig, line 111:
```
if err := ValidatePasswordProp(config.Password); err != nil {
  er = append(er, err)
}
```

src/lib/config/config-validator.go
```diff
@@ -1,61 +1,131 @@
...
+		}
+		if !Method[route.Method] {
+			routesErr = append(routesErr, fmt.Errorf("for \"%s\" \"%s\" is not a valid http method, Update the Method", route.URL, route.Method))
+		}
+	}
+	return routesErr
 }
 
 // Validate validates the local configuration file.
 func (inst *Config) Validate() bool {
```
aquibbaig, line 105: I think we can return early here (if logging everything is not so important to us). If the first test fails we do not have to wait for other tests appending into the `errors` variable. Let's return a true at the end so that this function returns `isValid` only when all tests pass. Let me know what you think.

src/lib/config/config-validator.go
```diff
@@ -1,61 +1,131 @@
...
+	er := []error{}
```
aquibbaig, line 106: If we consider the above comment we probably do not need this error object. We can directly log the error and return a `boolean`.

aquibbaig (changes requested):
Looks good, just some minor fixes and we're good to go, thanks!

src/lib/config/config-validator.go
Sanketwable: My thinking over here was. If user gets his first error he will fix it and again run the server.
At step two he will find second error again he will follow similar approach. And so on.
So printing all the errors in the config file at once will help him fix all the error instantly. Without making him run the server multiple times.

aquibbaig: Fair point! Let's keep it as it is then.

aquibbaig:
@Sanketwable Please edit the old commit using `git commit --amend` if the changes correspond to the same file/logic. "Updating the changes" as a commit message seems ambiguous to folks too, thanks!

src/lib/config/config-validator.go
Sanketwable: Thanks🙃

Issue #455, opened 2021-03-31
Test intervals accepts negative values


Sanketwable:
**Describe the bug**
In the update-interval endpoint negatives values are also accepted as argument and are updated into local-config.yml

**To Reproduce**
Steps to reproduce the behaviour:
1. Run the UI and input negative value in config->Scrape intervals

**Expected behaviour**
As it is time in sec it must be uint. but for now it is a string.
**Screenshots**
<img width="1440" alt="Screenshot 2021-03-31 at 11 17 14 PM" src="https://user-images.githubusercontent.com/43716242/113188702-0276c100-9278-11eb-87b1-0f40b0c43878.png">

aryan0078:
@Sanketwable i want to fix this

Anuragtech02:
@Sanketwable I think this issue can be fixed from the UI itself by preventing the input from accepting negative values.

Sanketwable:
@Anuragtech02 This will be a temporary fix. In future when we will need this API than it will again cause an issue. So what I think is, in the server side will change the type if the value from string to uint. The client side request should be modified to accept the positive value and pass this as a uint to the endpoint. This will solve end to end issue. Thanks

Anuragtech02:
> @Anuragtech02 This will be a temporary fix. In future when we will need this API than it will again cause an issue. So what I think is, in the server side will change the type if the value from string to uint. The client side request should be modified to accept the positive value and pass this as a uint to the endpoint. This will solve end to end issue. Thanks

Ohh yes, if the type is a string then it must be changed to int. Got it! Anyways can I contribute to the client-side?

Sanketwable:
> > @Anuragtech02 This will be a temporary fix. In future when we will need this API than it will again cause an issue. So what I think is, in the server side will change the type if the value from string to uint. The client side request should be modified to accept the positive value and pass this as a uint to the endpoint. This will solve end to end issue. Thanks
> 
> Ohh yes, if the type is a string then it must be changed to int. Got it! Anyways can I contribute to the client-side?

I am not contributing to client-side. You may ask Mentors for this.🙃

Anuragtech02:
@muskankhedia @Harkishen-Singh @aquibbaig I'd like to work on this issue for the client-side.

Harkishen-Singh:
@Anuragtech02 you can submit a client check for this. thank you.

PR #472, opened 2021-04-07
Fixed: test intervals accepts negative values

yoayushraj99:
issue: #455 

before the fix:

Screenshot from 2021-04-07 20-52-25

after the fix:

Screenshot

### DIFF
--- a/dashboard/v1.1/src/pages/Config/components/IntervalDetails.tsx
+++ b/dashboard/v1.1/src/pages/Config/components/IntervalDetails.tsx
@@ -1,14 +1,23 @@
 import React, { useState } from 'react';
 import { TextField, Button, Typography } from '@material-ui/core';
+import Alert from '@material-ui/lab/Alert';
 import { HOST_IP } from '../../../utils/types';
 
 const IntervalDetails = (props: any) => {
-  const [inputValue, setInputValue] = useState<string>(
+  const [inputValue, setInputValue] = useState<Number>(
     props.durationValue || ''
   );
 
+  const [alert, setAlert] = useState<boolean>(false);
+
   const handleIntervalOnChange = e => {
-    setInputValue(e.target.value);
+    const value = e.target.value;
+    if (value >= 0) {
+      setInputValue(value);
+      setAlert(false);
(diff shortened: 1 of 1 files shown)


Sanketwable:
@yoayushraj99 According to me the better approach should be:
1. Passing uint to backend instead of passing a string as a json payload 
` {
       "value":"3"
} ` should be replace by ` { "value":3}`.
2. By this approach we will be able to fix both backend and frontend part.
3. Only frontend fix seems to be a temporary approach.

yoayushraj99:
> @yoayushraj99 According to me the better approach should be:
> 
> 1. Passing uint to backend instead of passing a string as a json payload
>    `{ "value":"3" }` should be replace by ` { "value":3}`.
> 2. By this approach we will be able to fix both backend and frontend part.
> 3. Only frontend fix seems to be a temporary approach.

Hey @Sanketwable thnkx for the suggestion but I really don't know how to fix the server part. I'm fine with the client part but not with the server side part.😅

Sanketwable:
@yoayushraj99 If you can fix the issue in frontend. i.e In the request pass uint value. I can fix the backend part. I will create a PR for this and mention it here.
So this part will be fixed properly
Thanks

yoayushraj99:
> @yoayushraj99 If you can fix the issue in frontend. i.e In the request pass uint value. I can fix the backend part. I have created a PR for this #474
> So this part will be fixed properly. You can work on client fixing
> Thanks

Hey @Sanketwable  Typescript doesn't have uint type  because of that i am using number data type and then checking that the input value is positive or not if its positive then set the value.

Sanketwable:
> > @yoayushraj99 If you can fix the issue in frontend. i.e In the request pass uint value. I can fix the backend part. I have created a PR for this #474
> 
> > So this part will be fixed properly. You can work on client fixing
> 
> > Thanks
> 
> 
> 
> Hey @Sanketwable  Typescript doesn't have uint type  because of that i am using number data type and then checking that the input value is positive or not if its positive then set the value.

That's ok you can pass the positive Integer.

PR #476, opened 2021-04-09
Validation of Configuraion file issue fixed

### LINKED ISSUE #88: design a tool to check the the validity of configuration file

Sanketwable:
Fixes #88


Sanketwable:
The above PR will print all the errors in config file at once. So it will be easy for user to fix it. Rather printing one error at a time.

Issue #477, opened 2021-04-11
GetLabels has large response time.


Sanketwable:
When the endpoint `/get-labels` is hit it iterates over all the routes present in the config file. And checks for different labels.
This approach have O(no of routes ) time complexity. In case of production we will have to deal with lakhs of different routes. In such case it will take time in seconds to compute different labels.
### I tried doing this with more no of routes: 
I have calculated the time required between endpoint hit and response sent.
### The results are:
1.  It needs huge computation.
2. Takes more than **5 sec**.
<img width="1440" alt="GetLables" src="https://user-images.githubusercontent.com/43716242/114295262-c0444f80-9ac1-11eb-8495-c464f82a6cff.png">

Approach to perform this in O(1) time:
1.  We will store all the different labels in local-config. 
2.  Whenever new route is added we will cross check all the labels and add new label to local-config.
3.  At the time of endpoint `/get-label` hits it will only return the labels struct. It will not iterate over all the routes again.

Sanketwable:
@Harkishen-Singh @aquibbaig @muskankhedia can I work on this?

PR #481, opened 2021-04-12
OptimisationOfGetLabelsEndpoint

Sanketwable:
Fixes https://github.com/bench-routes/bench-routes/issues/477


Sanketwable:
We can further optimise this if we have label_ID for every label. This will then save memory and  storage. As we will not have to add every label to route. instead we can only add Label_IDs.

PR #491, opened 2021-05-08
Fix: QueryResponse not found

Rits1272:
Hello @Harkishen-Singh, it seems that whenever there is no block in the `Blockstream`, the error gets triggered at the `querier.go` as `FalseTypes`. So, I have simply added a check at the start of the function for checking the length of the `Blockstream` and the error seems to get resolved. Kindly check it

### DIFF
--- a/tsdb/querier/querier.go
+++ b/tsdb/querier/querier.go
@@ -103,6 +103,9 @@ func (q *Query) exec(blockStream []tsdb.Block, encoding bool) (i interface{}) {
 	// the total time involved in performing a particular query.
 	// This can be later benchmarked and compared with other algorithms.
 	base, stamp := getBaseResponse(q.Range)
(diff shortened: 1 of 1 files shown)


Tushar3099 (commented):
Hello @Rits1272, thanks for the PR. The solution you provided seems to work as it doesn't prompt any error but it's actually a workaround. Since we are returning before defer function, the code inside defer won't execute, hence the error is not popping. 
The problem is that we are returning the `[]byte` type in function `ReturnMessageResponse` in all cases whether `encoding` is true or false. The function should send `QueryResponse` in case of encoding is false and `[]byte` in case of encoding is true.

Rits1272:
Okay, @Tushar3099 I got your point, and thank you for the review. I will make the changes accordingly!

